PR #3, opened 2023-03-11
restore ci/cd pipeline - update action versions

mobidick1969:
Version up workflow action versions 

Motivation:
Current CI/CD pipelines are using obsolete action versions

Modification:
Change action versions

Result
CI/CD Pipeline works well


.github/workflows/mysql5-6.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - name: Shutdown the Default MySQL
         run: sudo service mysql stop
       - name: Set up MySQL 5.6
-        uses: mirromutth/mysql-action@v1.1
+        uses: jchrys/mysql-action@v1.1
```
jchrys: could we use asyncerio/**?

jchrys (changes requested):

Issue #39, opened 2023-03-20
Handling int<lenenc> values greater than 2^63-1 in MySQL OKPackets and related packets


jchrys:
## Issue Description

It has come to our attention that MySQL's OKPackets and other packets containing `int<lenenc>` types may potentially have values greater than 2^63-1. As per the MySQL documentation, [`int<lenenc>`](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_dt_integers.html#sect_protocol_basic_dt_int_le) supports such values.

## Current Implementation

Our current implementation does not take this into account, and we need to address this issue to ensure proper handling of `int<lenenc>` values exceeding 2^63-1.

## Impact on Real-world Usage

It is worth noting that this issue does not have a significant impact on real-world usage. Nonetheless, when converting the value to a string representation, it might be presented as a negative number.

## Proposed Solution

To enhance the accuracy and reliability of our implementation, we need to implement a patch that properly handles `int<lenenc>` values exceeding 2^63-1.

mirromutth:
This is caused by `VarIntUtils.readVarInt` trying to read 8 bytes and returning it directly as an integer for performance reasons, which should theoretically be wrapped with an object (such as writing an `UnsignedLong` or using a `BigInteger`).

However, this lower-level decoding operation may affect performance if using wrapper classes. Maybe add some code comments to inform the developer how the return value should be used instead of using a wrapper class.

jchrys:
Resolved by @mirromutth (https://github.com/asyncer-io/r2dbc-mysql/pull/161, #168 )

PR #78, opened 2023-04-05
Handling int<lenenc> fields from MySQL OK_Packets

svats0001:
Hi @jchrys ,

#39 

**Motivation:** Incoming values from MySQL OK_Packets that exceed 2^63-1 will incorrectly be represented as negative numbers in the string representation of the message object.

**Modifications:** I changed the toString method in the OkMessage class so that the unsigned long fields are converted into unsigned strings. I'm not aware of any other classes that deal with mySQL length encoded integer fields.

**Results:** When a value greater than 2^63-1 is received from a MySQL OK_Packet, that value is preserved in the string representation of the OkMessage object.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
@@ -111,14 +111,14 @@ public final class OkMessage implements WarningMessage, ServerStatusMessage, Com
     @Override
     public String toString() {
         if (warnings == 0) {
-            return "OkMessage{affectedRows=" + affectedRows + ", lastInsertId=" + lastInsertId +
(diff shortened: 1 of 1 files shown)


jchrys:
Hello, @svats0001.
This looks great to me.
Thanks very much.
if you have time could you make your commit message follow below format like other commit messages?


```
title

motivation:

modifications:

results:

```

jchrys:
Unfortunately, we're currently experiencing issues with our CI system when running tests on forked pull requests. Once this issue is resolved, I would be happy to merge your changes into our codebase.

svats0001:
No worries. I've updated my original commit message.

jchrys (approved):

jchrys:
Thanks a lot!

Issue #65, opened 2023-04-01
Support r2dbc-spi's extension(optional) interface 'Wrapped'


jchrys:
The Wrapped interface is an optional extension provided by the R2DBC specification to 
access an instance of a resource which has been wrapped and for implementers to expose wrapped resources. 

By implementing the Wrapped interface, vendors can expose their vendor-specific resources in a standard way. 

This feature is not mandatory to implement, but we would like to add support for it in our implementation. 

Refer to the R2DBC specification for more details: https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#wrapped.

PR #88, opened 2023-04-13
Added Wrapped interface

svats0001:
Hi @jchrys,

#65 

EDIT: The Wrapped interface already exists in the io.r2dbc.spi package, I think by support you meant the implementation?

I created the Wrapped optional interface. I'm unsure if I need to implement this anywhere or if it's entirely up to the end user. 

**Motivation:** To add support for the Wrapped optional interface in the R2DBC specification so that we can access an instance of a resource that has been wrapped or for implementers to expose wrapped resources.

**Modification:** Added the Wrapped interface into the io.asyncer.r2dbc.mysql package.

**Result:** We can now access an instance of a resource that has been wrapped or expose wrapped resources.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/Wrapped.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/Wrapped.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2023 asyncer.io projects
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


jchrys:
Hello, @svats0001.
Yes, You're correct. we need to implement it.
I think we can refer mysql-connector-j's implementation. (https://javadoc.io/doc/com.mysql/mysql-connector-j/latest/index.html)
However, as it's an optional feature and we have not received any requests from users for this functionality, we do not need to prioritize it immediately. That being said, if you are interested and have the bandwidth to work on this feature, we would certainly appreciate your efforts.

Thanks!

svats0001:
I'll have a look at implementing it when possible. I'll close this PR now.

PR #95, opened 2023-04-21
Deprecate SOCKET_TIMEOUT option

jchrys:
Motivation:
The SOCKET_TIMEOUT option has been identified as a NO-OP, meaning it has no effect on the underlying connection. To reduce confusion and potential issues for users, it is necessary to deprecate this option.

Modification:
Added a deprecation annotation to the SOCKET_TIMEOUT option and included a log warning message to inform users that the option is not supported and will be ignored.

Result:
The SOCKET_TIMEOUT option is now marked as deprecated, and users will be notified with a log warning message if they attempt to use it. This change encourages users to remove any references to SOCKET_TIMEOUT in their code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,7 +112,7 @@ ConnectionFactoryOptions options = ConnectionFactoryOptions.builder()
     .option(PASSWORD, "database-password-in-here") // optional, default null, null means has no password
     .option(DATABASE, "r2dbc") // optional, default null, null means not specifying the database
     .option(CONNECT_TIMEOUT, Duration.ofSeconds(3)) // optional, default null, null means no timeout
-    .option(Option.valueOf("socketTimeout"), Duration.ofSeconds(4)) // optional, default null, null means no timeout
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -112,7 +112,7 @@ ConnectionFactoryOptions options = ConnectionFactoryOptions.builder()
     .option(PASSWORD, "database-password-in-here") // optional, default null, null means has no password
     .option(DATABASE, "r2dbc") // optional, default null, null means not specifying the database
     .option(CONNECT_TIMEOUT, Duration.ofSeconds(3)) // optional, default null, null means no timeout
-    .option(Option.valueOf("socketTimeout"), Duration.ofSeconds(4)) // optional, default null, null means no timeout
+    .option(Option.valueOf("socketTimeout"), Duration.ofSeconds(4)) // deprecated since 1.0.1, since it has been a NO-OP and no longer has any functionality
```
jchrys:
```suggestion
    .option(Option.valueOf("socketTimeout"), Duration.ofSeconds(4)) // deprecated since 1.0.1, because it has no effect and serves no purpose.
```

Issue #76, opened 2023-04-04
[feature]Request to use Testcontainers for R2DBC MySQL Integration Testing


jchrys:
**Is your feature request related to a problem? Please describe.**
Currently, our integegration tests rely on docker-image that should be set up outside of this repository.

**Describe the solution you'd like**
Using testcontainers to run our integration tests.

**Additional context**

saurabhyadavdev:
Hey, I would like to contribute here. I am not able to see where are you using containers currently in code/docs. 

Is it here https://github.com/asyncer-io/r2dbc-mysql/blob/0f584e3dcaecf69f49badef5407dda325479aa20/src/test/java/io/asyncer/r2dbc/mysql/IntegrationTestSupport.java#L79 using mysql container ran externally and we need to run. using MysqlContainer in Integration tests?

jchrys:
Yes, that is correct. 
Currently, we're not using testcontainers in our integration tests, and we're setting up an external MySQL database using the `asyncer-io/mysql-action`. (you can refer `./github/workflows/mysql*.yml)
We want to remove this and use testcontainers instead to simplify the setup process and make it easier for contributors to run integration tests locally.

saurabhyadavdev:
Thanks, I can pick this task.

PR #109, opened 2023-04-24
Using Testcontainers for R2DBC MySQL Integration Testing #108

saurabhyadavdev:
Motivation:
To use testcontainers for R2DBC MySQL Integration Testing #76 

Modification:

1. Removing block to start/destroy mysql container from ci-integration-tests.yml
2. In IntegrationTestSupport adding static block to start mysql container. 
3. Use mapped port of container while building MySqlConnectionConfiguration

Result:
All tests working locally.


.github/workflows/ci-integration-tests.yml
```diff
@@ -19,13 +19,5 @@ jobs:
           distribution: temurin
           java-version: 8
           cache: maven
-      - name: Shutdown the Default MySQL
-        run: sudo service mysql stop
```
jchrys, line 23: Would you mind kindly sharing with us the reason for deleting this line? We appreciate your insights

src/test/java/io/asyncer/r2dbc/mysql/IntegrationTestSupport.java
```diff
@@ -38,10 +39,33 @@ abstract class IntegrationTestSupport {
...
     private final MySqlConnectionFactory connectionFactory;
 
+    private static final MySQLContainer<?> container;
+
     IntegrationTestSupport(MySqlConnectionConfiguration configuration) {
         this.connectionFactory = MySqlConnectionFactory.from(configuration);
     }
 
+    static {
+        String password = System.getProperty("test.mysql.password");
```
jchrys: I believe it's no longer necessary to retrieve the password from system properties. Could we consider removing this line and using a hardcoded password instead?

src/test/java/io/asyncer/r2dbc/mysql/IntegrationTestSupport.java
```diff
@@ -38,10 +39,33 @@ abstract class IntegrationTestSupport {
...
+        String version = System.getProperty("test.mysql.version");
```
jchrys: How about to use a fallback version(8), if the version is not explicitly specified?

jchrys (commented):
Thanks very much for your work!
Just small things.

jchrys:
close after making coauthored https://github.com/asyncer-io/r2dbc-mysql/pull/264 based off of this

Issue #66, opened 2023-04-01
Support opitonal Closeable interface for other stateful objects


jchrys:
While Connection already implements Closeable as a mandatory part of R2DBC, other stateful objects such as ConnectionFactory could benefit from providing a way to release their resources through implementing the Closeable interface. 

This is an optional feature, so it is not mandatory to implement it according to the R2DBC specification.

However, it would be useful for those who need to manually close their resources without blocking the caller.

refer: https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#closeable

JohnNiang:
IMO, the MySQLConnectionFactory isn't a stateful object event if it contains a client (Mono type).

But I'm willing to refactor it to make the method `MySqlConnectionFactory#create` clearer.

jchrys:
@JohnNiang 
Thanks for your thoughts, that make sense.
If you have a refactoring approach in mind to make `MySqlConnectionFactory#create` clearer, I’d love to see it!

PR #110, opened 2023-04-28
Added support for Closeable in MySqlConnectionFactory

svats0001:
Hi @jchrys,

#66

I added support for the Closeable interface in MySqlConnectionFactory and I also removed the invocation of the deprecated getSocketTimeout() and replaced it with null in MySqlConnectionFactory. Let me know if there are any problems in the code.

**Motivation:** MySqlConnectionFactory could benefit from a way to release its resources through implementing the Closeable interface.

**Modification:** MySqlConnectionFactory now implements Closeable and contains the close() method that will close the underlying MySqlConnection resource. I also added a logger to add the close start/end progress to the log. I removed the configuration.getSocketTimeout() parameter from Client.connect in the from() method and replaced it with null.

**Result:**  MySqlConnectionFactory now has a way for users to release its resources.


jchrys:
hello, @svats0001.
I think it seems that it creates a new connection and returns close Mono. (it doesn't seem to affect the factory)

svats0001:
Oh I see. I'm not too sure what's required to close the ConnectionFactory. I'll try again later. Closing the PR.

PR #111, opened 2023-05-03
Implemented Closeable in MySqlConnectionFactory

svats0001:
@jchrys #66 

I added support for the Closeable interface in MySqlConnectionFactory and I also removed the invocation of the deprecated getSocketTimeout() and replaced it with null in MySqlConnectionFactory. Let me know if there are any problems in the code.

Motivation: MySqlConnectionFactory could benefit from a way to release its resources through implementing the Closeable interface.

Modification: MySqlConnectionFactory now implements Closeable and contains the close() method that will remove the reference to the connection publisher. I removed the final keyword from the client reference so that it can be reassigned. I removed the configuration.getSocketTimeout() parameter from Client.connect in the from() method and replaced it with null.

Result: MySqlConnectionFactory now has a way for users to release its resources.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
@@ -26,6 +26,7 @@ import io.asyncer.r2dbc.mysql.constant.SslMode;
 import io.asyncer.r2dbc.mysql.extension.CodecRegistrar;
 import io.netty.buffer.ByteBufAllocator;
 import io.netty.channel.unix.DomainSocketAddress;
+import io.r2dbc.spi.Closeable;
 import io.r2dbc.spi.ConnectionFactory;
 import io.r2dbc.spi.ConnectionFactoryMetadata;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
```diff
@@ -108,6 +109,15 @@ public static MySqlConnectionFactory from(MySqlConnectionConfiguration configura
                 });
         }));
     }
+    
+    @Override
+    public Mono<Void> close() {
+    	return Mono.defer(() -> {
+    		this.client = null;
```
jchrys, line 116: Hello @svats0001,

I hope you're doing well. Unfortunately, I'm unable to accept this change at the moment, as it appears not to be thread-safe and there seems to be no handling for a null client. Would it be possible for us to postpone this change and perhaps discuss potential solutions to address these concerns?

Thank you for your understanding.

svats0001: Hi @jchrys,

Yep that makes sense. I'll make some changes and create another PR at a later date. I'll have a further read about thread safety but would a synchronized block on the client object work? I can add handling for the null client in the close method, is the create() method in MySqlConnectionFactory the only other place that I need to handle the null client?

PR #112, opened 2023-05-07
Support multi-host 

jchrys:
MySqlConnectionConfiguration
- [x] getHosts() returns `List<HostAndPort>`
- [x] getHaMode() returns parsed `HaMode`

MySqlConnectionConfiguration.Builder
- [x] port(int) sets default Port
- [x] host(String) supports multi-host
- [x] removed `isHost()`. (equivalent to  `unixSocket == null`)
- [x] field `domain` is removed and added `unixSocket` and `host` fields

TODO
- [ ] ADD MultiHostConnection or client that supports HaMode. (expected behavior: when it encounters an error propagate the error to all the task in the request queue and reconnect to appropriate host if needed)
- [ ] ADD Tests for above (Unit and Integration)
- [ ] Won't reconnect or try to recover if connection is explicitly closed

Refs
- mysql-connector-j spec. (https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-multi-host-connections.html)

Common Specs (some are from mysql-connector-j docs)
- Each multi-host connection is a wrapper of the underlying physical connections.
- Each of the underlying physical connections has its own session. Sessions cannot be tracked, shared, or copied, given the MySQL architecture.
- Every switch between physical connections means a switch between sessions.
- Within a transaction boundary, there are no switches between physical connections. Beyond a transaction boundary, there is no guarantee that a switch does not occur.

Fallback
connectionUrl
`r2dbc:pool:mysql:fallback://host:defaultPort,host2:port2/database?propertyName=propertyValue`

LoadBalance
connectionUrl
`r2dbc:pool:mysql:loadbalance://host:defaultPort,host2:port2/database?propertyName=propertyValue`

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/constant/HaMode.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/constant/HaMode.java
@@ -1,0 +1,40 @@
+package io.asyncer.r2dbc.mysql.constant;
+
+import org.jetbrains.annotations.Nullable;
+
+public enum HaMode {
+    NONE(HostsCardinality.SINGLE),
+    FALLBACK(HostsCardinality.MULTIPLE),
+    LOADBALANCE(HostsCardinality.ONE_OR_MORE),
+    ;
+
+    private final HostsCardinality hostsCardinality;
+
+    HaMode(HostsCardinality hostsCardinality) {
+        this.hostsCardinality = hostsCardinality;
+    }
+
+    public static HaMode parse(@Nullable String haMode, int hostCardinality) {
+        if (hostCardinality == 1) {
+           if (null == haMode || NONE.name().equalsIgnoreCase(haMode)) {
+               return NONE;
+           }
+        }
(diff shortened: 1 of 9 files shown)


src/main/java/io/asyncer/r2dbc/mysql/constant/HaMode.java
```diff
@@ -0,0 +1,40 @@
...
+        if (hostCardinality == 1) {
+           if (null == haMode || NONE.name().equalsIgnoreCase(haMode)) {
+               return NONE;
+           }
+        }
+        if (hostCardinality > 1) {
+            if (null == haMode || FALLBACK.name().equalsIgnoreCase(haMode)) {
+                return FALLBACK;
+            }
+            if (LOADBALANCE.name().equalsIgnoreCase(haMode)) {
```
jchrys, line 27: `Loadbalance`'s hostsCardinality is one_or_more but it doesn't support hostsCardinality == 1

Issue #113, opened 2023-05-08
[enhancement]replace ReactorNettyClient#Identity() with Function.identity()


jchrys:
We can remove `ReactorNettyClient#identity()`

DieMass:
Hi, @jchrys ! I would like to be assigned to this issue if possible

jchrys:
Nothing stops you. Thanks a lot!

PR #116, opened 2023-05-15
replace ReactorNettyClient#Identity() with Function.identity()

DieMass:
Motivation:
Closing [issue-113](https://github.com/asyncer-io/r2dbc-mysql/issues/113)

Modification:
Replaced the custom class ReactorNettyClient#Identity() with Function.identity()

Result:
Сode has become more universal

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
@@ -286,8 +286,6 @@ final class ReactorNettyClient implements Client {
         }
     }
 
-    @SuppressWarnings("unchecked")
-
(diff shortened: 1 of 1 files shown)


jchrys (approved):
@DieMass 
Thanks a lot!

src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
```diff
@@ -287,9 +287,6 @@ private void handleClose() {
     }
 
     @SuppressWarnings("unchecked")
```
jchrys: I don't think we need this line. could you check?

DieMass: fixed

PR #119, opened 2023-05-28
Fix request queue delaying

jchrys:
Motivation:
The request queue could contain cancelled tasks, which caused delays and stalls.

Modification:
Implemented Filtered out cancelled tasks from the queue.

Result:
The queue now functions better.
Resolves issue #114

PR #120, opened 2023-05-28
Drain request queue when connection is closed

jchrys:
Motivation:
Ensure that the request queue is drained when connection is closed, as there is a chance to remain unprocessed. This situation can prevent errors from being propagated effectively.

Modification:
Implemented proper draining of the request queue and sent an exit mesage.

Result:
The connection is now correctly closed, ensuring the proper handling of requests and preventing any unprocessed errors.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
@@ -200,14 +200,14 @@ final class ReactorNettyClient implements Client {
                 sink.success();
                 return;
             }
-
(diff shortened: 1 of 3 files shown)


src/main/java/io/asyncer/r2dbc/mysql/client/ReactorNettyClient.java
```diff
@@ -200,14 +200,14 @@ public Mono<Void> close() {
                 sink.success();
                 return;
             }
-
-            requestQueue.submit(RequestTask.wrap(sink, Mono.fromRunnable(() -> {
+            requestQueue.dispose();
```
jchrys, line 203: No need to dispose unless it has an error

jchrys:
resolved by #119

PR #157, opened 2023-10-07
Added Support for providing configuration option for supplying password function

bhosale:
Motivation:
Currently r2dbc-mysql does not support IAM based Authentication for authenticating with AWS Aurora RDS database. The way IAM based authentication works is requesting token from AWS RDS for that hostname and username (which is same as AWS IAM Role name). These tokens are valid for 15 minutes, cannot reuse same token after 15 minutes.
By adding configuration option for supplying password function, whenever a new connection is made then password is retrieved using supplier function each time.

Modification:
Modified `MySqlConnectionFactoryProvider` - Added new configurable option.
`Option<Supplier<Mono<String>>> PASSWORD_SUPPLIER = Option.valueOf("passwordSupplier");`
Modified `MySqlConnectionConfiguration` - Added the new configuration for Password Supplier function.
`Supplier<Mono<String>> passwordSupplier;`
Modified `MySqlConnectionFactory` - Retrieves Password Supplier function from configuration, and then retrieves password each time connection factory is created.
 

Result:
Users can provide a supplier function using `PASSWORD_SUPPLIER` option. This function will be used for retrieving password/token each time. 

```
public ConnectionFactory writeConnectionFactory(final RdsTokenGenerator rdsTokenGenerator) {
        return ConnectionFactories.get(ConnectionFactoryOptions.builder()
            .option(ConnectionFactoryOptions.DRIVER, "mysql")
            .option(ConnectionFactoryOptions.HOST, "Hostname of AWS Aurora DB instance")
            .option(ConnectionFactoryOptions.PORT, 3306)
            .option(ConnectionFactoryOptions.USER, "IAM ROLE Having access to RDS")
            .option(MySqlConnectionFactoryProvider.PASSWORD_SUPPLIER, rdsTokenGenerator::generateAuthenticationToken)
            .build());
    }
```

Example of `RdsTokenGenerator`
```
public class RdsTokenGenerator {
    public Mono<String> generateAuthenticationToken() {
        return Mono.fromCallable(() -> RdsUtilities.builder()
            .credentialsProvider(DefaultCredentialsProvider.create())
            .region(Region.US_EAST_1)
            .build();
            .generateAuthenticationToken((builder) ->
                builder
                    .hostname(hostname)
                    .port(port)
                    .username(user)
            ))
            .flatMap(token -> LOGGER.info("Retrieved token from RdsUtilities")
                .then(Mono.just(token)))
            .subscribeOn(Schedulers.boundedElastic());
    }

}
```

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
@@ -21,7 +21,7 @@ import io.asyncer.r2dbc.mysql.constant.ZeroDateOption;
 import io.asyncer.r2dbc.mysql.extension.Extension;
 import io.netty.handler.ssl.SslContextBuilder;
 import org.jetbrains.annotations.Nullable;
-import reactor.core.publisher.Mono;
+import org.reactivestreams.Publisher;
 
 import javax.net.ssl.HostnameVerifier;
 import java.net.Socket;
@@ -33,7 +33,6 @@ import java.util.Objects;
 import java.util.ServiceLoader;
 import java.util.function.Function;
 import java.util.function.Predicate;
-import java.util.function.Supplier;
 
 import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.require;
 import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.requireNonNull;
@@ -93,14 +92,14 @@ public final class MySqlConnectionConfiguration {
     private final int prepareCacheSize;
 
     private final Extensions extensions;
-    private final Supplier<Mono<String>> passwordSupplier;
+    private final Publisher<String> passwordSupplier;
 
     private MySqlConnectionConfiguration(boolean isHost, String domain, int port, MySqlSslConfiguration ssl,
         boolean tcpKeepAlive, boolean tcpNoDelay, @Nullable Duration connectTimeout,
         @Nullable Duration socketTimeout, ZeroDateOption zeroDateOption, @Nullable ZoneId serverZoneId,
         String user, @Nullable CharSequence password, @Nullable String database,
         @Nullable Predicate<String> preferPrepareStatement, int queryCacheSize, int prepareCacheSize,
-        Extensions extensions, @Nullable Supplier<Mono<String>> passwordSupplier) {
+        Extensions extensions, @Nullable Publisher<String> passwordSupplier) {
         this.isHost = isHost;
         this.domain = domain;
         this.port = port;
@@ -208,7 +207,7 @@ public final class MySqlConnectionConfiguration {
         return extensions;
     }
 
-    Supplier<Mono<String>> getPasswordSupplier() {
+    Publisher<String> getPasswordSupplier() {
         return passwordSupplier;
     }
 
@@ -336,7 +335,7 @@ public final class MySqlConnectionConfiguration {
         private final List<Extension> extensions = new ArrayList<>();
 
         @Nullable
-        private Supplier<Mono<String>> passwordSupplier;
+        private Publisher<String> passwordSupplier;
 
         /**
          * Builds an immutable {@link MySqlConnectionConfiguration} with current options.
@@ -795,7 +794,7 @@ public final class MySqlConnectionConfiguration {
          * @param passwordSupplier function to retrieve password before making connection.
          * @return this {@link Builder}.
          */
-        public Builder passwordSupplier(Supplier<Mono<String>> passwordSupplier) {
+        public Builder passwordSupplier(Publisher<String> passwordSupplier) {
             this.passwordSupplier = passwordSupplier;
             return this;
         }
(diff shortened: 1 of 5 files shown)


jchrys:
Hello, @bhosale thank you so much for your efforts in supporting this.

I believe we could utilize `Supplier<String>` instead of `Supplier<Mono<String>>` to avoid imposing the adoption of Reactor and simplify the interface. WDYT?

Additionally, could you please write a test case for this change?

jchrys:
PTAL @JohnNiang @mobidick1969

bhosale:
Thanks @jchrys for reviewing and providing feedback on the pull request. I have added couple of unit test cases.

As I understand the connection is established each time an SQL statement is executed, if we use just `Supplier<String>` instead of `Supplier<Mono<String>>`, there's a potential risk that the supplier function might involve blocking operations, which could have adverse consequences.

jchrys:
Establishing a connection involves multiple round-trip operations, which is why many of our users choose to pool connections using r2dbc-pool or similar solutions (and we highly recommend it).

I understand your point. I believe it would be more suitable to expose `org.reactivestreams.Publisher<String>` instead of `Supplier<Mono<String>>`. This way, we provide a more general interface that accommodates users who may use RxJava, Reactor, or other reactive libraries

bhosale:
Agree, I am also using pool. 

`Publisher<String>` makes sense instead of `Subscriber<Mono<String>>`. Updated the code with that change.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -204,6 +207,10 @@ Extensions getExtensions() {
         return extensions;
     }
 
+    Publisher<String> getPasswordSupplier() {
```
jchrys, line 213: `@Nullable`?

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -91,13 +92,14 @@ public final class MySqlConnectionConfiguration {
     private final int prepareCacheSize;
 
     private final Extensions extensions;
+    private final Publisher<String> passwordSupplier;
```
jchrys, line 97: seems `@Nullable` is missing

jchrys: +Please update MySqlConnectionConfiguration#toString,#equals,#hash as well.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java
```diff
@@ -23,12 +23,14 @@
...
 import io.r2dbc.spi.ConnectionFactoryProvider;
 import io.r2dbc.spi.Option;
+import org.reactivestreams.Publisher;
 
 import javax.net.ssl.HostnameVerifier;
 import java.time.Duration;
 import java.time.ZoneId;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
```
jchrys: I think we can remove this

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
```diff
@@ -91,22 +94,54 @@ public static MySqlConnectionFactory from(MySqlConnectionConfiguration configura
...
+            final MySqlSslConfiguration ssl,
+            final SocketAddress address,
+            final String database,
+            final String user,
+            final SslMode sslMode,
+            final ConnectionContext context,
+            final Extensions extensions,
+            final Predicate<String> prepare,
+            final int prepareCacheSize,
+            final CharSequence password) {
```
jchrys: `@Nullable`

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
```diff
@@ -91,22 +94,54 @@ public static MySqlConnectionFactory from(MySqlConnectionConfiguration configura
...
+            final MySqlConnectionConfiguration configuration,
+            final LazyQueryCache queryCache,
+            final MySqlSslConfiguration ssl,
+            final SocketAddress address,
+            final String database,
+            final String user,
+            final SslMode sslMode,
+            final ConnectionContext context,
+            final Extensions extensions,
+            final Predicate<String> prepare,
```
jchrys: `@Nullable`

jchrys (changes requested):
Just a nit

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
bhosale: Addressed

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
bhosale: Addressed

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java
bhosale: Addressed

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
bhosale: Addressed

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
bhosale: Addressed

jchrys (approved):
@bhosale Thanks a lot!

PR #164, opened 2023-12-15
Upgrade dependencies for 0.9.x

mirromutth:
Motivation:

Upgrade outdated dependencies for `0.9.x`.

It also solve a security warning for `logback-classic`.

Modification:

`pom.xml`

Result:

All dependencies will be upgraded except Reactor and dependencies that do not support Java 8. The Reactor should be upgraded carefully.


jchrys (approved):
lgtm

PR #161, opened 2023-12-13
Correct unsigned long handling

mirromutth:
Motivation:

Length-encoded integers should be handled correctly, especially those larger than `Long.MAX_VALUE`. See also #39. 

Modification:

- Add code comments to explain how to handle the length encoded integer.
- Correct `LargeFieldReader.readSlice()` method, it should use the length as an unsigned long.

Result:

- `LargeFieldReader.readSlice()` should now work correctly for fields those size is greater than 2<sup>63</sup>-1

PR #168, opened 2023-12-18
Correct unsigned long handling (#161)

jchrys:
Motivation:

Length-encoded integers should be handled correctly, especially those larger than `Long.MAX_VALUE`. See also #39.

Modification:

- Add code comments to explain how to handle the length encoded integer.
- Correct `LargeFieldReader.readSlice()` method, it should use the length as an unsigned long.

Result:

- `LargeFieldReader.readSlice()` should now work correctly for fields those size is greater than 2<sup>63</sup>-1


mirromutth (approved):
LGTM

Issue #158, opened 2023-11-18
[feature] create database if not exists


trajano:
**Is your feature request related to a problem? Please describe.**
Because R2DBC kicks in before Liquibase and the mysql driver doesn't support auto creation, it fails creating the connection as the database does not exist

**Describe the solution you'd like**
Same as JDBC MySQL createDatabaseIfNotExists=true property

**Additional context**

mirromutth:
This feature involves modifications to `MySqlConnection.init` and `QueryFlow.clientCapability`.

If the `createDatabaseIfNotExists` is enabled, then `disableConnectWithDatabase` whatever `database` is set or not. Then `HandshakeResponse.encode` will not send `database` to server on handshake. After handshake succeed, send `CREATE DATABASE IF NOT EXISTS` query and [`COM_INIT_DB`](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_init_db.html) command if the `createDatabaseIfNotExists` is enabled.

One thing that can be optimized here is that we can try to send only `COM_INIT_DB` first, and then if it fails, try to send `CREATE` statement and `COM_INIT_DB` again.

This feature can also be enhanced for those server versions that do not support initializing the database over the handshake protocol.

jchrys:
#162

PR #169, opened 2023-12-18
Add option to enable `createDatabaseIfNotExist`

jchrys:
Motivation:

See also #158 .

It should be named `createDatabaseIfNotExist`, the same as JDBC naming.

Modification:

- [x] Add option to enable `createDatabaseIfNotExist`
- [x] Support `COM_INIT_DB` and database can be initialized after handshake
- [x] Support sending `CREATE DATABASE IF NOT EXISTS` statement and retrying `COM_INIT_DB` after `COM_INIT_DB` fails for the first time

Result:

Support `createDatabaseIfNotExist`.


mirromutth (commented):
LGTM

mirromutth (approved):

Issue #64, opened 2023-04-01
Support for r2dbc-spi's Optional Interface 'LifeCycle'


jchrys:
While implementing r2dbc-spi, the interface 'LifeCycle' is optional and not mandated by the specification. 
However, we believe it is important to provide support for this interface as it can benefit minority use cases. 
You can refer to this link for more information: https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#lifecycle.

svats0001:
Hi @jchrys,

Do these interfaces need to be implemented anywhere or do they just need to be created? I think MySqlConnection has to implement Lifecycle. I don't think Wrapped needs to be implemented anywhere, it's up to the end user. Closeable might have to be implemented by MySqlConnectionFactory etc.

jchrys:
You're absolutely right. There is no need to implement the Wrapped<T> interface for MySqlConnection and MySqlConnectionFactory, as they are not wrappers. If, however, we come across a class that wraps an r2dbc-spi implementation, it would be advisable to implement Wrapped<T> in that case.

mirromutth:
The `MySqlConnection` should to do nothing in `postAllocate`, only contains one job in `preRelease`, which is to rollback if it is in a transaction.

Which means `postAllocate` will return an empty `Mono`, and `preRelease` will be an alias as `rollbackTransaction`.

mirromutth:
Hmmm, I noticed that the `COM_RESET_CONNECTION` command, which should be more suitable as `preRelease`, it resets almost all state, including temporary tables, transactions, etc.

mirromutth:
Forget it, it turns out that `COM_RESET_CONNECTION` is not a good solution.

This command not being supported in 5.6.x and below, it will also cause all prepared statements to be cleared.

jchrys:
#171 #175

PR #171, opened 2023-12-20
Implement `Lifecycle` for `MySqlConnection`

mirromutth:
Motivation:

Implement `Lifecycle` for `MySqlConnection`. See also #64 

Modification:

The `MySqlConnection`.

Result:

The `MySqlConnection` implements `Lifecycle` and will rollback in `preRelease`.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
@@ -31,6 +31,7 @@ import io.netty.util.internal.logging.InternalLogger;
 import io.netty.util.internal.logging.InternalLoggerFactory;
 import io.r2dbc.spi.Connection;
 import io.r2dbc.spi.IsolationLevel;
+import io.r2dbc.spi.Lifecycle;
 import io.r2dbc.spi.TransactionDefinition;
 import io.r2dbc.spi.ValidationDepth;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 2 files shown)


jchrys (approved):
LGTM

mirromutth:
I re-implement it with `COM_RESET_CONNECTION`, which will reset all state, not just rollback transaction. That should make more sense.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
```diff
@@ -92,6 +94,19 @@ public final class MySqlConnection implements Connection, ConnectionState {
...
+    private static final BiConsumer<ServerMessage, SynchronousSink<Void>> RESET = (message, sink) -> {
+        if (message instanceof ErrorMessage) {
+            ErrorMessage msg = (ErrorMessage) message;
+            logger.warn("Reset connection failed: [{}] [{}] {}", msg.getCode(), msg.getSqlState(),
+                msg.getMessage());
+            sink.complete();
+        } else if (message instanceof CompleteMessage && ((CompleteMessage) message).isDone()) {
+            sink.complete();
+        } else {
+            ReferenceCountUtil.safeRelease(message);
```
JohnNiang: Do we forget to call `sink#complete()` after this line? Otherwise, if it enters this condition, it will never contimue. If I have overlooked anything, please feel free to correct me.

mirromutth: This command should only respond Ok/Err packets, so we wait for Ok/Err packets, and discard other packets at this line. Which means that if this line is reached, it is not yet complete.

See also [`MySqlConnection.PING`](https://github.com/asyncer-io/r2dbc-mysql/blob/trunk/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java#L80)

JohnNiang: Make sense.

mirromutth:
Looks like MySQL 5.6.x and below does not support `COM_RESET_CONNECTION`.

Then I found that the reset status of `COM_RESET_CONNECTION` seemed a bit unexpected. It will clean up all prepared statements (server-preparing), which will leave our `PrepareCache` in an error state.

So I reset the commit, sorry for that.

jchrys (approved):
lgtm

PR #173, opened 2023-12-20
Remove Synchronized Statements(Loom-friendly)

jchrys:
Motivation:
Synchronized statements tend to pin virtual threads to carrier threads.

Modification:
Replaced synchronized statements with ReentrantLock

Result:
Loom-friendly

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
@@ -35,6 +35,7 @@ import reactor.core.publisher.Mono;
 import java.net.InetSocketAddress;
 import java.net.SocketAddress;
 import java.util.Objects;
+import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.Predicate;
 
 import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.requireNonNull;
@@ -157,6 +158,8 @@ public final class MySqlConnectionFactory implements ConnectionFactory {
 
         private final int capacity;
 
+        private final ReentrantLock lock = new ReentrantLock();
+
         @Nullable
         private volatile QueryCache cache;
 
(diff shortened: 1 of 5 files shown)


src/main/java/io/asyncer/r2dbc/mysql/cache/PrepareBoundedCache.java
```diff
@@ -33,6 +34,8 @@ final class PrepareBoundedCache extends HashMap<String, Lru.Node<Integer>> imple
 
     private final Lru<Integer> protection;
 
+    private final ReentrantLock reentrantLock = new ReentrantLock();
```
mirromutth: Maybe it should be named `lock`, the same as the other classes

mirromutth:
After this PR is merged, we can consider removing `RequestQueue.requireDisposed` and `RequestQueue.disposed`. This is probably over-designed.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
```diff
@@ -167,11 +170,14 @@ private LazyQueryCache(int capacity) {
         public QueryCache get() {
             QueryCache cache = this.cache;
             if (cache == null) {
-                synchronized (this) {
+                try {
+                    lock.lock();
```
JohnNiang: How about moving `lock#lock()` before try statement? See the example at [here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html).

mirromutth: @JohnNiang is right, all `lock()` should be preceded by `try`

mirromutth (changes requested):
`Lock#lock()` should be executed before the `try` statement.

> How about moving `lock#lock()` before try statement? See the example at [here](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/locks/ReentrantLock.html).

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactory.java
jchrys: Oh, I've just fixed it.
Thanks :D

src/main/java/io/asyncer/r2dbc/mysql/cache/PrepareBoundedCache.java
jchrys: fixed it

mirromutth (approved):
LGTM

PR #179, opened 2023-12-21
Enhance Transactional Methods for Improved Request Queue Handling

jchrys:
Motivation:
The current Transaction Methods implementation lacks consideration for the request queue, resulting in potential undefined behavior and issues.

Modifications:
Transaction Methods properly respect the request queue.

Result:
Bug resolved.
Enhanced stability and reliability.


mirromutth (approved):
The current code does cause some problems and we should fix it.

I was thinking that transaction state is necessary, like [ConnectionImpl#rollback()](https://github.com/mysql/mysql-connector-j/blob/release/8.x/src/main/user-impl/java/com/mysql/cj/jdbc/ConnectionImpl.java#L1798) does, it even throws an exception if connection is not in a transaction.

Maybe we should check the state in `RequestQueue` instead, I created a ticket #183 for that. I think we can merge this PR first for fix the bug, and consider how to improve state checking (including other states).

PR #180, opened 2023-12-21
Enhance Transactional Methods for Improved Request Queue Handling

jchrys:
Motivation:
The current Transaction Methods implementation lacks consideration for the request queue, resulting in potential undefined behavior and issues.

Modifications:
Transaction Methods properly respect the request queue.

Result:
Bug resolved.
Enhanced stability and reliability.


mirromutth (approved):

mirromutth:
Maybe we should check the state in `RequestQueue` instead, I created a ticket #183 for that.

I think we can merge this PR first for fix the bug, and consider how to improve state checking (including other states).

PR #185, opened 2023-12-29
Transaction States Should Be Checked In Queue

### LINKED ISSUE #183: [bug] States should be checked in queue
**Describe the bug**

Currently, `MySqlConnection` checks transaction state between the `Mono.defer` is subscribed and `Exchangeable` is executed.

It may cause some undefined behavior. All states checking will have same problem, like `MySqlConnection.isSessionAutoCommit`.

I was thinking that transaction state is necessary, like [ConnectionImpl#rollback()](https://github.com/mysql/mysql-connector-j/blob/release/8.x/src/main/user-impl/java/com/mysql/cj/jdbc/ConnectionImpl.java#L1798) does, it even throws an exception if connection is not in a transaction.

Firstly, we should remove transaction state checking in `MySqlConnection`, see also #179 does.

Then we can consider adding proper status checks in `Exchangeable` and `RequestQueue`.

**Expected behavior**

States should be checked in `RequestQueue` executing `Exchangeable` instead.

jchrys:
Motivation:
Currently, `MySqlConnection` checks state between the `Mono.defer` is subscribed and `Exchangeable` is executed. It may cause undefined behavior.

Modification:
Checks transaction state when request queue executes task.

Result:
Resolves #183

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
@@ -249,9 +249,9 @@ final class QueryFlow {
      */
     static Mono<Void> beginTransaction(Client client, ConnectionState state, boolean batchSupported,
         TransactionDefinition definition) {
-        StartTransactionState startState = StartTransactionState.of(state, definition);
+        final StartTransactionState startState = new StartTransactionState(state, definition);
 
-        if (batchSupported || startState.isSimple()) {
+        if (batchSupported) {
             return client.exchange(new TransactionBatchExchangeable(startState)).then();
         }
 
@@ -265,19 +265,25 @@ final class QueryFlow {
      * @param client          the {@link Client} to exchange messages with.
      * @param state           the connection state for checks and resets transaction statuses.
      * @param commit          if commit, otherwise rollback.
-     * @param lockWaitTimeout the lock wait timeout of the initial connection state.
      * @param batchSupported  if connection supports batch query.
      * @return receives complete signal.
      */
-    static Mono<Void> doneTransaction(Client client, ConnectionState state, boolean commit,
-        long lockWaitTimeout, boolean batchSupported) {
-        CommitRollbackState commitState = CommitRollbackState.of(state, commit, lockWaitTimeout);
+    static Mono<Void> doneTransaction(Client client, ConnectionState state, boolean commit, boolean batchSupported) {
+        final CommitRollbackState commitState = new CommitRollbackState(state, commit);
 
-        if (batchSupported || commitState.isSimple()) {
+        if (batchSupported) {
             return client.exchange(new TransactionBatchExchangeable(commitState)).then();
         }
 
         return client.exchange(new TransactionMultiExchangeable(commitState)).then();
+    }
+
+    static Mono<Void> createSavepoint(Client client, ConnectionState state, String name, boolean batchSupported) {
+        final CreateSavepointState savepointState = new CreateSavepointState(state, name);
+        if (batchSupported) {
+            return client.exchange(new TransactionBatchExchangeable(savepointState)).then();
+        }
+        return client.exchange(new TransactionMultiExchangeable(savepointState)).then();
     }
 
     /**
@@ -979,28 +985,22 @@ abstract class AbstractTransactionState {
 
     final ConnectionState state;
 
+    final List<String> statements = new ArrayList<>(5);
+
     /**
      * A bitmap of unfinished tasks, the lowest one bit is current task.
      */
     int tasks;
 
-    private final List<String> statements;
-
     @Nullable
     private String sql;
 
-    protected AbstractTransactionState(ConnectionState state, int tasks, List<String> statements) {
+    protected AbstractTransactionState(ConnectionState state) {
         this.state = state;
-        this.tasks = tasks;
-        this.statements = statements;
     }
 
     final void setSql(String sql) {
         this.sql = sql;
-    }
-
-    final boolean isSimple() {
-        return statements.size() == 1;
     }
 
     final String batchStatement() {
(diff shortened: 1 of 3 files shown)


src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -265,19 +266,25 @@ static Mono<Void> beginTransaction(Client client, ConnectionState state, boolean
...
+        if (batchSupported) {
+            return client.exchange(new TransactionBatchExchangeable(commitStateSupplier)).then();
         }
 
-        return client.exchange(new TransactionMultiExchangeable(commitState)).then();
+        return client.exchange(new TransactionMultiExchangeable(commitStateSupplier)).then();
+    }
+
+    static Mono<Void> createSavepoint(Client client, ConnectionState state, String name, boolean batchSupported) {
+        Supplier<AbstractTransactionState> commitStateSupplier = () -> CreateSavepointState.of(state, name);
```
mirromutth: Maybe it should be `savepointState`

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1227,16 +1234,63 @@ private static String buildStartTransaction(TransactionDefinition definition) {
...
+
+    private static final int CREATE_SAVEPOINT = 2;
+
+    private CreateSavepointState(ConnectionState state, int tasks, List<String> statements) {
+        super(state, tasks, statements);
+    }
+
+    @Override
+    boolean cancelTasks() {
+        return false;
```
mirromutth, line 1236: If it looks like following:

```java
@Override
boolean cancelTasks() {
    this.tasks = state.isInTransaction() ? CREATE_SAVEPOINT : START_TRANSACTION | CREATE_SAVEPOINT;

    return false;
}
```

Then `TransactionBatchExchangeable` or `TransactionMultiExchangeable` will check transaction state at `subscribe`, and `this.tasks` will be changed to the correct.

This avoids the use of `@Nullable state`, additional `Supplier`, and `assert`.

Or is there something else to consider?

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1227,16 +1234,63 @@ private static String buildStartTransaction(TransactionDefinition definition) {
...
+    @Nullable
+    private AbstractTransactionState state;
+
+    TransactionBatchExchangeable(Supplier<AbstractTransactionState> stateSupplier) {
+        this.stateSupplier = stateSupplier;
     }
 
     @Override
     public void accept(ServerMessage message, SynchronousSink<Void> sink) {
+        assert state != null;
```
mirromutth: Maybe consider to use `cancelTasks` for avoid `Nullable` and `assert`?

mirromutth:
Hmm, the naming of the method `cancelTasks` may hard to understand. Maybe we can consider changing it to `recalculateTasks` or `recalculateOrCancelTasks`

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
jchrys: Nice feedback.  Let me change it :D

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
jchrys: done

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -72,9 +72,11 @@
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
```
mirromutth: It should be unused import

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1075,26 +1089,18 @@ protected boolean process(int task, SynchronousSink<Void> sink) {
...
-            List<String> statements = new ArrayList<>(2);
-
-            statements.add("SET innodb_lock_wait_timeout=" + lockWaitTimeout);
-            statements.add(doneSql);
-
-            return new CommitRollbackState(state, LOCK_WAIT_TIMEOUT | COMMIT_OR_ROLLBACK, statements);
-        }
-
-        return new CommitRollbackState(state, COMMIT_OR_ROLLBACK, Collections.singletonList(doneSql));
+    static CommitRollbackState of(ConnectionState state, boolean commit) {
```
mirromutth: Maybe just use constructor instead of a static method, because we calculate statements in `cancelTasks` now.

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1106,25 +1112,37 @@ final class StartTransactionState extends AbstractTransactionState {
 
     private static final int START_TRANSACTION = 4;
 
-    private final long lockWaitTimeout;
-
-    @Nullable
-    private final IsolationLevel isolationLevel;
+    private static final int CANCEL = 8;
```
mirromutth, line 1109: I'm not sure if the `CANCEL` flag is necessary. Currently, if `cancelTasks` returns `true`, that means "cancel the current request/exchange", then `CANCEL` flag will not be actually used.

So I think we can just remove the flag, and use `0` for the bitmap value if it should be cancelled.

Or if the `CANCEL` flag should be used, we can change `cancelTasks` to return `void`, then make `TransactionBatchExchangeable#subscribe` and `TransactionMultiExchangeable#subscribe` check the calculated flags instead of the return value.

mirromutth (commented):
Good idea, it makes sense to calculate statements and task bitmaps in same method.

I think it's ready to be merged for now, but maybe it could use a little more polish?

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
jchrys: The current implementation sends a `PING` client message when `cancelTasks` returns `true` and processes a `ServerMessage` with the `START_TRANSACTION` task flag turned on. I just separated the flag into `START_TRANSACTION` and `CANCEL` to make flags more clear.

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
jchrys: Sure, Thanks!

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
jchrys: done

jchrys:
> Good idea, it makes sense to calculate statements and task bitmaps in same method.
> 
> I think it's ready to be merged for now, but maybe it could use a little more polish?

Refined the code to enhance clarity and reduced the reliance on `AbstractTransactionState`. 
Please review, and please let me know if want to revert to the latest version.

jchrys:
> Good idea, it makes sense to calculate statements and task bitmaps in same method.
> 
> I think it's ready to be merged for now, but maybe it could use a little more polish?

Sure, I think we could address the known issue first and then work on further polishing it.
@mirromutth @JohnNiang PTAL

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -983,26 +989,20 @@ abstract class AbstractTransactionState {
      * A bitmap of unfinished tasks, the lowest one bit is current task.
      */
     int tasks;
-
-    private final List<String> statements;
+    List<String> statements = new ArrayList<>(5);
```
jchrys: final

Issue #176, opened 2023-12-21
Support recognition of MariaDB


mirromutth:
MariaDB contains more flags in the last 4 bytes of the 10 bytes filter (the reserved field) of `HandshakeV10Request`, which expands capabilities to 64 bits.

The `CLIENT_MYSQL=1` flag is used to indicate whether to use the MySQL compatible protocol. MariaDB 10.2 and above will unset this flag. If this flag is unset, the server is assumed to be MariaDB.

Note that in the MySQL server, this flag is called `LONG_PASSWORD`, and it is not currently used by `r2dbc-mysql`. We can consider reinterpreting it for support MariaDB.

See also https://mariadb.com/kb/en/connection/#capabilities

mirromutth:
Done in PR

PR #186, opened 2024-01-04
Add MariaDB capabilities support

mirromutth:
Motivation:

See also #176 .

Modification:

Add recognition for MariaDB, and extend `Capability` to 64-bits.

Result:

Extend `Capability` to 64-bits, and extended bits will be read and sent at handshake.

Currently, the driver will unset all MariaDB features, we can enable them after we support these features.


jchrys (approved):
lgtm

PR #187, opened 2024-01-08
Add MariaDB capabilities support

jchrys:
Motivation:

See also https://github.com/asyncer-io/r2dbc-mysql/issues/176 .

Modification:

Add recognition for MariaDB, and extend Capability to 64-bits.

Result:

Extend Capability to 64-bits, and extended bits will be read and sent at handshake.

Currently, the driver will unset all MariaDB features, we can enable them after we support these features.


jchrys:
PTAL @JohnNiang , @mirromutth

JohnNiang (approved):
LGTM

PR #201, opened 2024-01-16
Add `RETURNING` support for MariaDB 10.5.1+

### LINKED ISSUE #177: Support `RETURNING` for MariaDB
MariaDB 10.5.1+ support to use `INSERT ... RETURNING col1[, col2[, col3...]]` for returns a result-set of the inserted rows. The response looks like the execution result of a `SELECT` statement, and the last `OkMessage` will NOT contain the updated rows.

It will change the logic of `MySqlStatement.returnGeneratedValues` in MariaDB 10.5.1+ .

See also https://mariadb.com/kb/en/insertreturning/

mirromutth:
Motivation:

Add `RETURNING` clause support for MariaDB 10.5.1+ .

See also #177 .

Modification:

Add `RETURNING` clause in `MySqlStatement#returnGeneratedValues` and query message packets.

Result:

For MariaDB 10.5.1 and above, `MySqlStatement#returnGeneratedValues` will use `RETURNING` clause instead of `LAST_INSERT_ID`.

Notice:

For MariaDB 10.5.0 and below, `MySqlStatement#returnGeneratedValues` will still use `LAST_INSERT_ID` due to these versions do not support `RETURNING` clause.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlResult.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlResult.java
@@ -40,6 +40,7 @@ import reactor.core.publisher.Flux;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.SynchronousSink;
 
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
 import java.util.function.Function;
@@ -227,34 +228,35 @@ public final class MySqlResult implements Result {
 
     private static class MySqlUpdateCount implements UpdateCount {
 
-        protected final OkMessage message;
+        protected final long rows;
 
-        private MySqlUpdateCount(OkMessage message) {
-            this.message = message;
-        }
+        private MySqlUpdateCount(long rows) { this.rows = rows; }
 
         @Override
         public long value() {
-            return message.getAffectedRows();
+            return rows;
         }
     }
 
     private static final class MySqlOkSegment extends MySqlUpdateCount implements RowSegment {
 
+        private final long lastInsertId;
+
         private final Codecs codecs;
 
         private final String keyName;
 
-        private MySqlOkSegment(OkMessage message, Codecs codecs, String keyName) {
-            super(message);
+        private MySqlOkSegment(long rows, long lastInsertId, Codecs codecs, String keyName) {
+            super(rows);
 
+            this.lastInsertId = lastInsertId;
             this.codecs = codecs;
             this.keyName = keyName;
         }
 
         @Override
         public Row row() {
-            return new InsertSyntheticRow(codecs, keyName, message.getLastInsertId());
+            return new InsertSyntheticRow(codecs, keyName, lastInsertId);
         }
     }
 
@@ -269,6 +271,8 @@ public final class MySqlResult implements Result {
         @Nullable
         private final String syntheticKeyName;
 
+        private final AtomicLong rowCount = new AtomicLong(0);
+
         private MySqlRowMetadata rowMetadata;
 
         private MySqlSegments(boolean binary, Codecs codecs, ConnectionContext context,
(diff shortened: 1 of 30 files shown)


mirromutth:
@jchrys

Currently, it is simplify to add `RETURNING` following original SQL statement without checking.

Perhaps the statement parser needs to be rewritten to support checking whether the statement supports `RETURNING` (`INSERT`/`REPLACE`/`DELETE`) or already includes `RETURNING`, like [`r2dbc-mariadb` does](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/master/src/main/java/org/mariadb/r2dbc/MariadbClientParameterizedQueryStatement.java#L126)?

jchrys:
> @jchrys
> 
> Currently, it is simplify to add `RETURNING` following original SQL statement without checking.
> 
> Perhaps the statement parser needs to be rewritten to support checking whether the statement supports `RETURNING` (`INSERT`/`REPLACE`/`DELETE`) or already includes `RETURNING`, like [`r2dbc-mariadb` does](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/master/src/main/java/org/mariadb/r2dbc/MariadbClientParameterizedQueryStatement.java#L126)?

I think  the current change is suitable for merging, as MariaDB will naturally return a syntax error in case of incorrect `RETURNING` usage. I agree that enhancing the parser, akin to what's done in `r2dbc-mariadb`, better be next step.

src/main/java/io/asyncer/r2dbc/mysql/MySqlResult.java
```diff
@@ -309,10 +315,17 @@ public void accept(ServerMessage message, SynchronousSink<Segment> sink) {
 
                 this.rowMetadata = MySqlRowMetadata.create(metadataMessages);
             } else if (message instanceof OkMessage) {
-                Segment segment = generatedKeyName == null ? new MySqlUpdateCount((OkMessage) message) :
-                    new MySqlOkSegment((OkMessage) message, codecs, generatedKeyName);
+                OkMessage msg = (OkMessage) message;
+
+                if (MySqlStatementSupport.supportReturning(context) && msg.isEndOfRows()) {
+                    sink.next(new MySqlUpdateCount(rowCount.get()));
```
jchrys: Do might need to add `rowCount.set(0)` after this line?

jchrys (approved):
lgtm 👍

mirromutth:
Columns should not be quoted, it can be an expression. e.g. `INSERT INTO ... RETURNING CURRENT_TIMESTAMP`.

If we quoted it, like in the example above it would be <code>INSERT INTO ... RETURNING &#96;CURRENT_TIMESTAMP&#96;</code>, which would be confusing. User wants the result of the `CURRENT_TIMESTAMP` expression, not a column named `CURRENT_TIMESTAMP`.

Keep columns raw and let ORM/QueryBuilder handle it.

mirromutth:
@jchrys It is foreseeable that this PR will change the behavior of `returnGenerateValues` when connecting to MariaDB 10.5.1+, maybe we should merge it with the next minor version?

For example, if a user is using `returnGenerateValues` without arguments, we will return a `Result` with one and only one row, which contains a `LAST_INSERT_ID` column. This modification will change the number of result rows (maybe more, maybe 0 by `INSERT ... SELECT ... RETURNING ...`), and the row metadata will contain all data columns in the table.

jchrys:
> @jchrys It is foreseeable that this PR will change the behavior of `returnGenerateValues` when connecting to MariaDB 10.5.1+, maybe we should merge it with the next minor version?
> 
> For example, if a user is using `returnGenerateValues` without arguments, we will return a `Result` with one and only one row, which contains a `LAST_INSERT_ID` column. This modification will change the number of result rows (maybe more, maybe 0 by `INSERT ... SELECT ... RETURNING ...`), and the row metadata will contain all data columns in the table.

I agree. 😄

jchrys (approved):

Issue #192, opened 2024-01-11
Aligning `Connection#setTransactionIsolationLevel` Behavior with `r2dbc-spi` Specification


jchrys:
# Issue

The current implementation of `Connection#setTransactionIsolationLevel` does not align with its specification. According to the specification(), this method should configure the isolation level for the connection, and the specified isolation level should apply to all subsequent transactions. However, the current behavior applies the isolation level only to the next transaction.

# Proposal
Aligning the behavior of `Connection#setTransactionIsolationLevel` with its specification. The implementation should be adjusted so that the configured isolation level persists for all subsequent transactions until explicitly changed.

# Action Items
- Modify the implementation of `Connection#setTransactionIsolationLevel` to align with the specification.(`AbstractTransactionState` could be utilized if needed )
- Verify and update the transaction isolation behavior in the entire project to ensure consistency and adherence to the proposed changes.

## Refs
- r2dbc-spi specification: https://github.com/r2dbc/r2dbc-spi/blob/160cd2743f8ce8551618abcd96fa1d7ee37f8f38/r2dbc-spi/src/main/java/io/r2dbc/spi/Connection.java#L221C12-L221C12
- mysql-dev doc: https://dev.mysql.com/doc/refman/8.0/en/set-transaction.html#set-transaction-scope

JohnNiang:
I'm willing to contribute this enhancement.

/assign

jchrys:
> I'm willing to contribute this enhancement.
> 
> /assign

Thanks!

jchrys:
#202 resolved this issue

PR #202, opened 2024-01-16
Ensure isolation level is applied to subsequent transactions after setting

JohnNiang:
Motivation:

Aligning Connection#setTransactionIsolationLevel Behavior with r2dbc-spi Specification. See https://github.com/asyncer-io/r2dbc-mysql/issues/192 for more.

Modification:

Change session isolation level as well after invoking `Connection#setTransactionIsolationLevel`.

Result:

All subsequent transactions will be applied the isolation level set by `Connection#setTransactionIsolationLevel`.

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
@@ -129,7 +129,7 @@ public final class MySqlConnection implements Connection, Lifecycle, ConnectionS
 
     private final MySqlConnectionMetadata metadata;
 
-    private final IsolationLevel sessionLevel;
+    private volatile IsolationLevel sessionLevel;
 
     private final QueryCache queryCache;
 
@@ -303,13 +303,27 @@ public final class MySqlConnection implements Connection, Lifecycle, ConnectionS
         return currentLevel;
     }
 
+    /**
+     * Gets session transaction isolation level(Only for testing).
+     *
+     * @return session transaction isolation level.
+     */
+    IsolationLevel getSessionTransactionIsolationLevel() {
+        return sessionLevel;
+    }
+
     @Override
     public Mono<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
         requireNonNull(isolationLevel, "isolationLevel must not be null");
 
-        // Set next transaction isolation level.
-        return QueryFlow.executeVoid(client, "SET TRANSACTION ISOLATION LEVEL " + isolationLevel.asSql())
-            .doOnSuccess(ignored -> setIsolationLevel(isolationLevel));
+        // Set subsequent transaction isolation level.
+        return QueryFlow.executeVoid(client, "SET SESSION TRANSACTION ISOLATION LEVEL " + isolationLevel.asSql())
+            .doOnSuccess(ignored -> {
+                this.sessionLevel = isolationLevel;
+                if (!this.isInTransaction()) {
+                    this.currentLevel = isolationLevel;
+                }
+            });
     }
 
     @Override
--- a/src/main/java/io/asyncer/r2dbc/mysql/message/client/TextQueryMessage.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/message/client/TextQueryMessage.java
@@ -19,6 +19,7 @@ package io.asyncer.r2dbc.mysql.message.client;
 import io.asyncer.r2dbc.mysql.ConnectionContext;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import java.util.Objects;
 import reactor.core.publisher.Mono;
 
 import java.nio.charset.Charset;
@@ -65,6 +66,23 @@ public final class TextQueryMessage implements ClientMessage {
                 throw e;
             }
         });
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) {
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+        TextQueryMessage that = (TextQueryMessage) o;
+        return Objects.equals(sql, that.sql);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(sql);
     }
 
     @Override
--- a/src/test/java/io/asyncer/r2dbc/mysql/ConnectionIntegrationTest.java
+++ b/src/test/java/io/asyncer/r2dbc/mysql/ConnectionIntegrationTest.java
@@ -140,6 +140,51 @@ class ConnectionIntegrationTest extends IntegrationTestSupport {
     }
 
     @Test
+    void setTransactionLevelNotInTransaction() {
+        complete(connection ->
+            // check initial session isolation level
+            Mono.fromSupplier(connection::getTransactionIsolationLevel)
+                .doOnSuccess(it -> assertThat(it).isEqualTo(REPEATABLE_READ))
+                .then(connection.beginTransaction())
+                .doOnSuccess(ignored -> assertThat(connection.isInTransaction()).isTrue())
+                .then(Mono.fromSupplier(connection::getTransactionIsolationLevel))
+                .doOnSuccess(it -> assertThat(it).isEqualTo(REPEATABLE_READ))
+                .then(connection.rollbackTransaction())
+                .then(connection.setTransactionIsolationLevel(READ_COMMITTED))
+                // ensure that session isolation level is changed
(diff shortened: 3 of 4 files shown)


src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
```diff
@@ -303,13 +303,25 @@ public IsolationLevel getTransactionIsolationLevel() {
...
+    IsolationLevel getSessionTransactionIsolationLevel() {
+        return sessionLevel;
+    }
+
     @Override
     public Mono<Void> setTransactionIsolationLevel(IsolationLevel isolationLevel) {
         requireNonNull(isolationLevel, "isolationLevel must not be null");
 
         // Set next transaction isolation level.
         return QueryFlow.executeVoid(client, "SET TRANSACTION ISOLATION LEVEL " + isolationLevel.asSql())
```
jchrys: I believe it is necessary to utilize the `SET SESSION TRANSACTION ISOLATION LEVEL` statement to apply this isolation level to all subsequent transactions.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
```diff
@@ -303,13 +303,25 @@ public IsolationLevel getTransactionIsolationLevel() {
...
-            .doOnSuccess(ignored -> setIsolationLevel(isolationLevel));
+            .doOnSuccess(ignored -> {
+                this.sessionLevel = isolationLevel;
+                resetIsolationLevel();
```
jchrys: Can modify the current isolation level only when there is no ongoing transaction. In MySQL, it is not permitted to change the isolation level of a transaction that is already in progress.

To accomplish this functionality, I suggest referring to `AbstractTransactionState`.

jchrys (changes requested):
just a nit

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
JohnNiang: You are right. The current isolation level will be reset automatically when commiting or rollbacking, please see https://github.com/asyncer-io/r2dbc-mysql/blob/96c11dcfe86e9a95cc857a5bfadc7906be5ef953/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java#L1088

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
JohnNiang: Added a corresponding unit test for `SET SESSION TRANSACTION ISOLATION LEVEL`.

src/test/java/io/asyncer/r2dbc/mysql/ConnectionIntegrationTest.java
```diff
@@ -139,6 +139,31 @@ void transactionDefinitionIsolationLevel() {
             }));
     }
 
+    @Test
```
jchrys, line 142: I'm looking to validate that `connection.getTransactionIsolationLevel()` correctly returns the current transaction isolation level. I believe the following test cases should effectively demonstrate this.  WDYT?

```
    @Test
    void setTransactionLevelNotInTransaction() {
        complete(connection ->
                // check initial session isolation level
                         Mono.just(connection.getTransactionIsolationLevel())
                             .doOnSuccess(it -> assertThat(it).isEqualTo(REPEATABLE_READ))
                             .then(connection.beginTransaction())
                             .doOnSuccess(ignored -> assertThat(connection.isInTransaction()).isTrue())
                             .then(Mono.just(connection.getTransactionIsolationLevel()))
                             .doOnSuccess(it -> assertThat(it).isEqualTo(REPEATABLE_READ))
                             .thenMany(Flux.merge(
                                     connection.rollbackTransaction(),
                                     // set transaction isolation level in no transaction
                                     connection.setTransactionIsolationLevel(READ_COMMITTED)
                             ))
                             // ensure that session isolation level is changed
                             .then(Mono.just(connection.getTransactionIsolationLevel()))
                             .doOnSuccess(it -> assertThat(it).isEqualTo(READ_COMMITTED))
                            .then(connection.beginTransaction())
                            .doOnSuccess(ignored -> assertThat(connection.isInTransaction()).isTrue())
                             // ensure transaction isolation level applies to subsequent transactions
                            .then(Mono.just(connection.getTransactionIsolationLevel()))
                            .doOnSuccess(it -> assertThat(it).isEqualTo(READ_COMMITTED))
        );
    }

    @Test
    void setTransactionLevelInTransaction() {
        complete(connection ->
                // check initial session transaction isolation level
                         Mono.just(connection.getTransactionIsolationLevel())
                             .doOnSuccess(it -> assertThat(it).isEqualTo(REPEATABLE_READ))
                             .thenMany(Flux.merge(
                                     connection.beginTransaction(),
                                     // set transaction isolation level in transaction
                                     connection.setTransactionIsolationLevel(READ_COMMITTED)
                             ))
                             // ensure that current transaction isolation level is not changed
                             .then(Mono.just(connection.getTransactionIsolationLevel()))
                             .doOnSuccess(it -> assertThat(it).isNotEqualTo(READ_COMMITTED))
                             .then(connection.rollbackTransaction())
                             .doOnSuccess(ignored -> assertThat(connection.isInTransaction()).isFalse())
                             // ensure that session isolation level is changed after rollback
                             .then(Mono.just(connection.getTransactionIsolationLevel()))
                             .doOnSuccess(it -> assertThat(it).isEqualTo(READ_COMMITTED))
                             // ensure transaction isolation level applies to subsequent transactions
                             .then(connection.beginTransaction())
                             .doOnSuccess(ignored -> assertThat(connection.isInTransaction()).isTrue())
        );
    }
 ```

ref 
from the mysql dev doc
```
Transaction characteristic scope

With the SESSION keyword:

The statement applies to all subsequent transactions performed within the current session.

The statement is permitted within transactions, but does not affect the current ongoing transaction.

If executed between transactions, the statement overrides any preceding statement that sets the next-transaction value of the named characteristics.
```

File: src/main/java/io/asyncer/r2dbc/mysql/message/client/TextQueryMessage.java [hunk 4]
```diff
@@ -67,6 +68,23 @@ public Mono<ByteBuf> encode(ByteBufAllocator allocator, ConnectionContext contex
         });
     }
 
+    @Override
```
jchrys, line 71: nice catch 👍

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnection.java
```diff
@@ -303,13 +303,27 @@ public IsolationLevel getTransactionIsolationLevel() {
...
         requireNonNull(isolationLevel, "isolationLevel must not be null");
 
-        // Set next transaction isolation level.
-        return QueryFlow.executeVoid(client, "SET TRANSACTION ISOLATION LEVEL " + isolationLevel.asSql())
-            .doOnSuccess(ignored -> setIsolationLevel(isolationLevel));
+        // Set subsequent transaction isolation level.
+        return QueryFlow.executeVoid(client, "SET SESSION TRANSACTION ISOLATION LEVEL " + isolationLevel.asSql())
+            .doOnSuccess(ignored -> {
+                this.sessionLevel = isolationLevel;
+                if (!this.isInTransaction()) {
```
JohnNiang, line 323: I'm not sure if the condtion is correct to set current level even if we reset isolation level when committing or rolling back.

jchrys: @mirromutth, Could you please take a look?

IMHO, It seems accurate since that `sessionLevel` denotes the isolation level for upcoming and subsequent transactions. 
When in a transaction, `currentLevel` indicates the isolation level of the current transaction. when not in a transaction, `currentLevel` represents the isolation level for the very next transaction.

JohnNiang: Maybe I should populate the current isolation level at the begining of beginTransaction().

jchrys: I believe the current implementation effectively deals with that scenario:

Case 1: When the isolation level is set within a transaction, the currentLevel will equal the sessionLevel following a rollback or commit. (Note: r2dbc-mysql doesn't allow a double begin.)
Case 2: If the isolation level is set outside of a transaction, this method will adjust both the session level and the current level.

As a result, in both scenarios, when you are not in a transaction(or before executing `beginTransaction()`), the condition currentLevel == sessionLevel is maintained.

jchrys (approved):
lgtm

mirromutth (approved):
LGTM

Issue #203, opened 2024-01-16
Check and remove MariaDB 10.x fake prefix in version


mirromutth:
MariaDB 10.x has a replication hack that sends a prefixed version `5.5.5-` in the handshake. e.g. `5.5.5-10.6.6-MariaDB`

Check the replication hack and remove fake prefix in `ServerVersion#parse`.

Consider adding a boolean property to `ServerVersion` to check if the server is MariaDB instead of just checking via capabilities.

See also [`r2dbc-mariadb` does](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/4f9385fd9cabcab4557c1f228cfdafa4a3a27ce2/src/main/java/org/mariadb/r2dbc/message/server/InitialHandshakePacket.java#L107)

jchrys:
for future ref
- https://jira.mariadb.org/browse/MDEV-4088

PR #204, opened 2024-01-16
Add support for MariaDB version pattern

mirromutth:
Motivation:

Should remove fake prefix `5.5.5-` for MariaDB 10.x

See also #203 

This is why #201 cannot assert that the server supports `RETURNING` clause.

Modification:

- Check and remove MariaDB version fake prefix
- Add `isMariaDb` to `ConnectionContext`, it will check  either capabilities or server version is present as MariaDB
- Correct server version checking, like TLS version, `@@transaction_isolation`/`@@tx_isolation`

Result:

Driver correctly parses and checks MariaDB server version


jchrys (approved):
Nice work! LGTM

PR #207, opened 2024-01-19
Fix NPE issue when batch execution is called without any queries

T45K:
Motivation:
(Please describe the problem you are trying to solve, or the new feature you are trying to add.)

we're using this R2DBC client with jOOQ.
NPE occurs when we give empty list to `DSLContext#batch`

```kotlin
val queries = emptyList()
dslContext.batch(queries).awaitLast()
```

this is because `StringBuilder` in `MySqlBatchingBatch` is not initialized until `add` method is called.
so, `execute` method will refer to not-initialized `builder` when `add` method is not called.

Modification:
(Please describe the changes you have made to the codebase, including any new files, modified files, or deleted files.)

~~modify `execute` to return `Flux#empty` when ` builder` is not initialized (i.e., `add` method is not called).~~
send empty string to client without NPE

Result:
(Please describe the expected outcome of your changes, and any potential side effects or drawbacks. 
If possible, please also include any relevant testing results.)

it now does not throw NPE ~~but returns empty flux instead.~~

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/MySqlBatchingBatch.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/MySqlBatchingBatch.java
@@ -20,6 +20,8 @@ import io.asyncer.r2dbc.mysql.client.Client;
 import io.asyncer.r2dbc.mysql.codec.Codecs;
 import reactor.core.publisher.Flux;
 
+import java.util.StringJoiner;
+
 import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.requireNonNull;
 
 /**
@@ -34,7 +36,7 @@ final class MySqlBatchingBatch extends MySqlBatch {
 
     private final ConnectionContext context;
 
-    private StringBuilder builder;
+    private final StringJoiner queries = new StringJoiner(";");
 
     MySqlBatchingBatch(Client client, Codecs codecs, ConnectionContext context) {
         this.client = requireNonNull(client, "client must not be null");
@@ -50,9 +52,9 @@ final class MySqlBatchingBatch extends MySqlBatch {
 
         if (index >= 0 && sql.charAt(index) == ';') {
             // Skip last ';' and whitespaces that following last ';'.
-            requireBuilder().append(sql, 0, index);
+            queries.add(sql.substring(0, index));
         } else {
-            requireBuilder().append(sql);
+            queries.add(sql);
         }
 
         return this;
@@ -60,10 +62,6 @@ final class MySqlBatchingBatch extends MySqlBatch {
 
     @Override
     public Flux<MySqlResult> execute() {
-        if (builder == null) {
-            return Flux.empty();
-        }
-
         return QueryFlow.execute(client, getSql())
             .map(messages -> MySqlResult.toResult(false, codecs, context, null, messages));
     }
@@ -79,15 +77,7 @@ final class MySqlBatchingBatch extends MySqlBatch {
      * @return current batching SQL statement
      */
     String getSql() {
-        return builder.toString();
-    }
-
-    private StringBuilder requireBuilder() {
-        if (builder == null) {
-            return (builder = new StringBuilder());
-        }
-
-        return builder.append(';');
+        return queries.toString();
     }
 
     private static int lastNonWhitespace(String sql) {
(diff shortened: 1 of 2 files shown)


mirromutth:
The [`io.r2dbc.spi.Batch` spec](https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#batches.executing) shows a specification like following:

> `Batch` objects are run by calling the execute() method after adding one or more SQL statements to a `Batch`.

So we cannot return an empty `Flux`. Here are some suggestions:

- Execute an empty query. Then server may return an error (e.g. MySQL and MariaDB), or do nothing (e.g. TiDB)
- Emit an exception that should not be a NPE, and it cannot be thrown directly.
  - In `r2dbc-mariadb` [MariadbBatch.java#L92](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/1.1.4/src/main/java/org/mariadb/r2dbc/MariadbBatch.java#L92), it will emit a `NoSuchElementException` due to `iterator.next()` without checking
  - Note: the empty checking should be deferred.

However, it seems like a code smell coming from `MySqlBatchingBatch.builder`, could you please add a `@Nullable` to the `builder` property? Then, we can see a null-check warning at `MySqlBatchingBatch.getSql()`, so we can add a null-check to the `getSql()` instead of `execute()`, which should make sense.

T45K:
thank you for confirming my PR!

> The [io.r2dbc.spi.Batch spec](https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#batches.executing) shows a specification like following:

> Batch objects are run by calling the execute() method after adding one or more SQL statements to a Batch.

sorry, i didn't know the spi spec :pray:

> Execute an empty query. Then server may return an error (e.g. MySQL and MariaDB), or do nothing (e.g. TiDB)

i think this is more reasonable. i'd like to rewrite in this way.

___
(question)

i think we can make `builder` not-null by initializing it when `MySqlBatchingBatch` is constructed.

```diff
-     private StringBuilder builder;
+     private final StringBuilder builder = new StringBuilder();
```

and then, `requireBuilder` can be rewritten like

```diff
    private StringBuilder requireBuilder() {
-      if (builder == null) {
-          return (builder = new StringBuilder());
-      }

+        if (builder.length() == 0) {
+            return builder;
+        }

        return builder.append(';');
    }
```

this looks better for me. what do you think?

mirromutth:
> i think we can make `builder` not-null by initializing it when `MySqlBatchingBatch` is constructed.

Hmmm, if we check only it is empty or not, consider following code:

```java
batch1.add(";").add("SELECT 1").execute()
batch2.add("").add("SELECT 1").execute()
batch3.add("SELECT 1").execute()
```

There `Batch`s will execute same query `SELECT 1`, but first and second `Batch`s should be `;SELECT 1`, third should be `SELECT 1`.

Maybe it would be simpler to refactor the entire `builder` into a `List`. Note that statement joins should be deferred.

T45K:
i see. i'll update this PR later

src/main/java/io/asyncer/r2dbc/mysql/MySqlBatchingBatch.java
```diff
@@ -34,7 +36,7 @@ final class MySqlBatchingBatch extends MySqlBatch {
 
     private final ConnectionContext context;
 
-    private StringBuilder builder;
+    private final StringJoiner queries = new StringJoiner(";");
```
T45K, line 39: i think `StringJoiner` is sufficient and simple enough in this use case.
so i used it instead of `List`

src/main/java/io/asyncer/r2dbc/mysql/MySqlBatchingBatch.java
```diff
@@ -75,15 +77,7 @@ public String toString() {
      * @return current batching SQL statement
      */
     String getSql() {
-        return builder.toString();
-    }
-
-    private StringBuilder requireBuilder() {
```
T45K, line 81: no longer necessary

T45K (commented):
@mirromutth 
i updated this PR. could you check this again?

mirromutth (approved):
LGTM

jchrys (approved):

PR #215, opened 2024-01-23
Add `LOCAL INFILE` support

### LINKED ISSUE #206: [feature] Support `LOCAL INFILE`

Currently, we have `Capability.LOCAL_FILES` marked as disabled, and it is not actually supported. The statement like following:

```sql
LOAD DATA LOCAL INFILE '/path/to/file' INTO TABLE `my_table`; 
```

After executing this statement (without error), the database server will immediately request the file content from the driver.

Note: MySQL 8.0+ has `@@global.local_infiles` disabled for security reasons, so this feature should be disabled by default.

- Add a `LocalInfileRequest` to package `message.server`, the header is `0xFB`
- Add a `LocalInfileResponse` to package `message.client`, each such message contains a file chunk or an empty chunk (means End-Of-File)
- Add `LOCAL INFILE` support in message decoder
- Add `LOCAL INFILE` support in `QueryFlow`, which should send file chunks and an empty chunk after receiving `LocalInfileRequest`
- Maybe add an option for specifying the max size of file chunk

**Additional context**

See also [MySQL](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_com_query_response_local_infile_request.html) and [MariaDB](https://mariadb.com/kb/en/packet_local_infile) docs

It is only supported in text protocol queries.

Reference from: mirromutth/r2dbc-mysql#110

mirromutth:
Motivation:

Add `LOCAL INFILE` support.

See also #206 .

Modification:

- [x] Improve envelope codec for `SubsequenceClientMessage`
- [x] Add `LocalInfileRequest`/`LocalInfileResponse` for text protocol
- [x] Add an option to enable/disable `LOCAL INFILE` support, and it should be a string to specify "safety path"
- [x] Add an option to specify file buffer size
- [x] Write docs

Result:

Query `LOCAL INFILE` support.


mirromutth:
Hi, @jchrys @JohnNiang

In JDBC-MySQL, `allowLoadLocalInfileInPath` will work only if `allowLoadLocalInfile=false`. The `allowLoadLocalInfile` option means "allow `LOAD DATA LOCAL INFILE` and allow it to load file data from any path". It seems to be for historical reasons.

For example:

- `allowLoadLocalInfile=true` for **allowed** and unrestricted `LOCAL INFILE`
- `allowLoadLocalInfileInPath=/opt` for **allowed** and **restricted** `LOCAL INFILE`
- `allowLoadLocalInfile=true&allowLoadLocalInfileInPath=/opt` for **allowed** and **UNRESTRICTED** `LOCAL INFILE`, `allowLoadLocalInfileInPath` will be ignored

In my opinion, `r2dbc-mysql` should make `allowLoadLocalInfile` just mean "allow `LOAD DATA LOCAL INFILE`", and that can be restricted by `loadLocalInfileInPath`. Which means, `loadLocalInfileInPath` will work only if `allowLoadLocalInfile=true`, and all `LOCAL INFILE` can only read files in the path.

For example:

- `allowLoadLocalInfile=true` for **allowed** and unrestricted `LOCAL INFILE`
- `allowLoadLocalInfile=true&loadLocalInfileInPath=/opt` for **allowed** and **restricted** `LOCAL INFILE`
- `loadLocalInfileInPath=/opt`, `LOCAL INFILE` will not be allowed

This prevents users from accidentally setting up unrestricted file access.

Or we can make it the same as JDBC.

Any ideas?

jchrys:
I agree with your idea but would like to suggest a slightly different approach.

What if we consider not supporting the `allowLoadLocalInfile` property? (or keep `(jdbc version of)allowLoadLocalInfile` = `false`)

I think that the property `allowLoadLocalInfileInPath={path}` could effectively convey the user's intentions(`allowLoadLocalInfileInPath=/` for unrestricted case as well).

What are your thoughts on this?

src/main/java/io/asyncer/r2dbc/mysql/message/client/LocalInfileResponse.java
```diff
@@ -0,0 +1,121 @@
...
+        return Flux.defer(() -> {
+            int bufferSize = context.getLocalInfileBufferSize();
+            AtomicReference<Throwable> error = new AtomicReference<>();
+
+            return Mono.<Path>create(sink -> {
+                try {
+                    Path safePath = context.getLocalInfilePath();
+                    Path file = Paths.get(this.path);
+
+                    if (safePath == null || file.startsWith(safePath)) {
```
jchrys: `safePath` is `null` means `allowLoadLocalInfileInPath = null`,  better not allow it when `safePath` is `null`

src/main/java/io/asyncer/r2dbc/mysql/internal/util/ReadCompletionHandler.java
```diff
@@ -0,0 +1,141 @@
...
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.nio.channels.AsynchronousFileChannel;
+import java.nio.channels.CompletionHandler;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * TODO: add javadoc here
```
jchrys: seems java doc is missing

jchrys (approved):
LGTM

Issue #190, opened 2024-01-11
Aligning `serverZoneId` Behavior with `connectionTimeZone`(mysql-connector-j)


jchrys:
# Issue

The current behavior in r2dbc-mysql when the `serverZoneId` property is not explicitly set is to use the server's timezone. However, this differs from the behavior of the `connectionTimeZone`(previously `serverTimeZone`) property in mysql-connector-j, causing inconsistency for users familiar with MySQL's JDBC driver.

# Proposal
I propose aligning the behavior of `serverZoneId` in r2dbc-mysql with the equivalent property in `mysql-connector-j`, 's `connectionTimeZone`. Specifically, when` serverZoneId` is not set, it should default to using the JVM's default timezone, similar to the behavior of `connectionTimeZone`.

# Rationale
Consistency across different MySQL drivers is crucial for users who may work with multiple technologies or migrate between them. Aligning r2dbc-mysql with the behavior of mysql-connector-j would reduce surprises for users and make it easier for them to switch between these drivers without encountering unexpected timezone-related issues. (e.g., https://github.com/micronaut-projects/micronaut-data/pull/2388)

# Refs
- https://dev.mysql.com/blog-archive/support-for-date-time-types-in-connector-j-8-0/

jchrys:
WDYT? @mirromutth

jchrys:
Hello @mirromutth, I would appreciate your input on this proposed change. Could you please share your thoughts? May I kindly join the discussion below, if you don't mind?). https://github.com/micronaut-projects/micronaut-data/pull/2388#discussion_r1482580836.

(additional ref: https://github.com/asyncer-io/r2dbc-mysql/issues/166)

mirromutth:
Currently, the default behavior should be same as `connectionTimeZone=SERVER` in JDBC MySQL.

If user set the `serverZoneId=X`, it should be same as `connectionTimeZone=X` in JDBC MySQL.

As I said in #166 , JDBC driver `ResultSet.getObject(int)` returns `LocalDateTime`, and the timezone of `LocalDateTime` will not be converted. Like following:

```java
// JVM is not in UTC

Class.forName("com.mysql.cj.jdbc.Driver");
String url = "jdbc:mysql://localhost:3306/r2dbc?connectionTimeZone=UTC";
String user = "root";
String password = "******";

try (Connection conn = DriverManager.getConnection(url, user, password);
     Statement stmt = conn.createStatement();
[...]

Issue #166, opened 2023-12-16
Discussion on Default Codecs for R2DBC MySQL


jchrys:
I'm initiating a discussion on a crucial matter concerning the default codec for `MySqlType.ID_TIMESTAMP` and `MySqlType.ID_DATETIME`.

Currently, we're using `ZonedDateTime.class` as default Codec for above MySqlTypes, which has resulted in compatibility issues with the Micronaut Data (R2DBC) project, as highlighted in this pull request (https://github.com/micronaut-projects/micronaut-data/pull/2388). 

Attempts to resolve this by setting `LocalDateTime.class` as the default codec introduced a backward compatibility problem with Spring Data, documented in the related issue (https://github.com/asyncer-io/r2dbc-mysql/issues/151), leading to a rollback to `ZonedDateTime.class` as the default (https://github.com/asyncer-io/r2dbc-mysql/pull/155).

I'm now contemplating setting `LocalDateTime.class` as the default codec after an upcoming minor version update(`1.1.0`/`0.10.0`).

(`LocalDateTime.class` was default in `dev.miku`'s, and `LocalDateTime.class` is default in latest `mariadb-r2dbc`.)

I need your opinions on how best to navigate these challenges. What do you think?
@mirromutth  @JohnNiang

mirromutth:
In fact, I have also hesitated for a long time about the default type at that time.

I also think `ZonedDateTime` makes more sense, but essentially the question is what does the ORM/QueryBuilder do after getting the result of `DATETIME`.

For example, if we are using JPA (3.2.0) + JDBC (8.2.0), create an entity with a `ZonedDateTime` property and try to `findAll`, we will see `com.mysql.cj.jdbc.result.ResultSetImpl.getTimestamp(int, Calendar)` (line 975) is called, it is not `ResultSet.getObject(int)` or `ResultSet.getObject(int, Class<T>)`. Which means, JPA will always get `java.sql.Timestamp` first, then convert it to `ZonedDateTime` after that. Actually, `mysql-connector-j` returns `LocalDateTime` as default type of `DATETIME` in `ResultSetImpl.getObject(int)`.

So `LocalDateTime` is the most common choice, but we also need more explanation and communication with the ORM/QueryBuilder framework maintainers to solve the overall problem.

PR #219, opened 2024-02-05
Add support for compression protocol

### LINKED ISSUE #216: [feature] Compression protocol support

Add support for compression protocol.

Capability flags:

- `Capability.COMPRESS`: use zlib for compression protocol
- `Capability.Z_STD_COMPRESSION`: use Z-standard

Enabling compression decreases performance, so its benefits occur primarily when there is low network bandwidth, network transfer time dominates the cost of compression and decompression operations, and result sets are large.

Add an option `compressionAlgorithms`, with default value `UNCOMPRESSED`, which auto selects the compression algorithm, prefer Z-standard.

Add compression packet codec before `MessageDuplexCodec`/`EnvelopeSlicer` if compression protocol is enabled.

**Additional context**

See also [MySQL](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_compression_packet.html) and [MariaDB](https://mariadb.com/kb/en/0-packet/#compressed-packet) docs.

Netty does not have a `ZstdDecoder` yet, see also PR netty/netty#13531. Use [zstd-jni](/luben/zstd-jni) instead, which is also used by `ZstdEncoder`. This dependency should be optional.

mirromutth:
Motivation:

See also #216 , [MySQL Connect Options --compression-algorithms](https://dev.mysql.com/doc/refman/8.0/en/connection-options.html#option_general_compression-algorithms)

Modification:

- Refactor message codecs, move sequence ID handling to `MessageDuplexCodec`
- Add compression codec and support for zstd and zlib
- Add configuration options for compression algorithms and zstd compression level
- Refactor the configuration way in the `IntegrationTestSupport`

I refactored codec handlers, used the original Sequence ID concept instead of "Envelope ID", there are two main reasons for this:

1. The `QueryFlow` should not consider how the packet should be encoded, the "envelope ID" is obviously something that the encoding layer needs to consider
   - This problem also occurred when implementing `LOCAL INFILE`. The `QueryFlow` needs to constructing `LocalInfileResponse` with  an envelope ID, but this is obviously not any logic about `LOCAL INFILE`
   - This also causes `MessageDecoder` to need to additionally consider whether to pass the envelope ID to `ServerMessage`
2. Compression protocol contains a compression sequence ID, it reset in the same time as the uncompressed sequence ID, but incremented independently.
   - If we support this like before, we need consider two envelope counter for uncompressed sequence ID and compression sequence ID at `QueryFlow`, all `ClientMessage`s and all `ServerMessage`s, which is bad.

The reason why this concept was created in the early days was that the relationship of sequence IDs between packets was not complicated at that time, and it was reset to 0 every time in the command phase.

Result:

User can configure the compression algorithm to enable the MySQL compression protocol and the driver will auto select the preferred algorithm to use. The compression protocol is disabled by default.


src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java
```diff
@@ -307,7 +347,7 @@ private static void setupHost(MySqlConnectionConfiguration.Builder builder, Opti
             .to(isSsl -> builder.sslMode(isSsl ? SslMode.REQUIRED : SslMode.DISABLED));
         mapper.optional(SSL_MODE).as(SslMode.class, id -> SslMode.valueOf(id.toUpperCase()))
             .to(builder::sslMode);
-        mapper.optional(TLS_VERSION).asStrings()
+        mapper.optional(TLS_VERSION).asArray(String[].class, Function.identity(), String[]::new)
```
jchrys, line 337: nice catch 👍

src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1027,13 +1060,31 @@ private HandshakeResponse createHandshakeResponse(int envelopeId, Capability cap
...
     }
 
     private static String authFails(String authType, String phase) {
         return "Authentication type '" + authType + "' must require SSL in " + phase + " phase";
     }
+
+    private static boolean isZstdSupported() {
+        try {
+            Class.forName("com.github.luben.zstd.Zstd", false,
+                LoginExchangeable.class.getClassLoader());
```
jchrys: Could you consider using `AccessController.doPrivileged` to obtain the classLoader, in case a SecurityManager is present?

jchrys (commented):
It's quite extensive, so it'll take me a bit longer to review thoroughly. Appreciate your patience!

src/main/java/io/asyncer/r2dbc/mysql/client/CompressionDuplexCodec.java
```diff
@@ -0,0 +1,243 @@
...
+
+            while (cumulated.readableBytes() >= Packets.MAX_PAYLOAD_SIZE) {
+                logger.trace("Accumulated to the maximum payload, compressing");
+
+                ByteBuf slice = cumulated.readSlice(Packets.MAX_PAYLOAD_SIZE);
+                ByteBuf compressed = compressor.compress(slice);
+
+                if (compressed.readableBytes() >= slice.readableBytes()) {
+                    logger.trace("Sending uncompressed due to compressed payload is larger than original");
+                    compressed.release();
```
jchrys, line 84: I've been testing whether sending uncompressed messages is feasible when the MySQL server has uncompressed connections disabled. It appears to result in an error. Let me  double-check this behavior. (ref: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_protocol_compression_algorithms)

mirromutth: Compression packets contain two schemas of packets: compressed and uncompressed.

Uncompressed packet is not a normal packet, it will contain compression header, normal header, and normal payload. Actually, the entire normal packet is a compression payload of uncompressed packet.

For example, this is a valid compression packet without compressing:

```pcap
0d 00 00 00 00 00 00 09    00 00 00 03 53 45 4c 45    ............SELE
43 54 20 32                                           CT 2
```

See also: https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_compression_packet.html#sect_protocol_basic_compression_packet_uncompressed_payload

jchrys: Thank you for clarifying that. 

I'll need to check since I've received an error message. It seems like a mistake in my configuration.

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -822,6 +853,64 @@ public Builder prepareCacheSize(int prepareCacheSize) {
...
+         * preferring zstd, then zlib. If the list does not contain {@link CompressionAlgorithm#UNCOMPRESSED}
+         * and the server does not support any algorithm in the list, an exception will be thrown when
+         * connecting.
+         * <p>
+         * Note: zstd requires a dependency {@code com.github.luben:zstd-jni}.
+         *
+         * @param compressionAlgorithms the list of compression algorithms.
+         * @return {@link Builder this}.
+         * @throws IllegalArgumentException if {@code compressionAlgorithms} is {@code null} or empty.
+         * @since 1.1.0
```
jchrys: maybe 1.1.2

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -822,6 +853,64 @@ public Builder prepareCacheSize(int prepareCacheSize) {
...
+         * <p>
+         * It is only used if zstd is chosen for the connection.
+         * <p>
+         * Note: MySQL protocol does not allow to set the zlib compression level of the server, only zstd is
+         * configurable.
+         *
+         * @param level the compression level.
+         * @return {@link Builder this}.
+         * @throws IllegalArgumentException if {@code level} is not between 1 and 22.
+         * @since 1.1.0
```
jchrys: 1.1.2

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java
```diff
@@ -181,9 +182,41 @@ public final class MySqlConnectionFactoryProvider implements ConnectionFactoryPr
...
+    /**
+     * Option to set compression algorithms.  Default to [{@link CompressionAlgorithm#UNCOMPRESSED}].
+     * <p>
+     * It will auto choose an algorithm that's contained in the list and supported by the server, preferring
+     * zstd, then zlib. If the list does not contain {@link CompressionAlgorithm#UNCOMPRESSED} and the server
+     * does not support any algorithm in the list, an exception will be thrown when connecting.
+     * <p>
+     * Note: zstd requires a dependency {@code com.github.luben:zstd-jni}.
+     *
+     * @since 1.1.0
```
jchrys: 1.1.2

src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java
```diff
@@ -181,9 +182,41 @@ public final class MySqlConnectionFactoryProvider implements ConnectionFactoryPr
...
+
+    /**
+     * Option to set the zstd compression level.  Default to {@code 3}.
+     * <p>
+     * It is only used if zstd is chosen for the connection.
+     * <p>
+     * Note: MySQL protocol does not allow to set the zlib compression level of the server, only zstd is
+     * configurable.
+     *
+     * @since 1.1.0
```
jchrys: 1.1.2

src/main/java/io/asyncer/r2dbc/mysql/message/client/ClientMessage.java
```diff
@@ -27,6 +27,19 @@
  */
 public interface ClientMessage {
 
+    default boolean isSequenceReset() {
```
jchrys, line 35: seems java doc is missing

jchrys (changes requested):
LGMT, just tiny details.

jchrys:
btw)
could you check below discussions once you have time?
https://github.com/asyncer-io/r2dbc-mysql/issues/190
https://github.com/asyncer-io/r2dbc-mysql/issues/166
above are related to micronaut's issue below.
https://github.com/micronaut-projects/micronaut-data/pull/2388

jchrys (approved):
LGTM

PR #220, opened 2024-02-05
Init global variable to GitHub Actions instead of test code

mirromutth:
Motivation:

Should not init MySQL global variables in test code.

Modification:

- Change the MySQL setup to use `docker compose` instead
- Add init options of `mysqld`
- Remove global variable initialization from the test code

Result:

Global variables will be initialized on GitHub Actions.


jchrys (approved):
LGTM

PR #230, opened 2024-02-13
Add support for decoding session state info

### LINKED ISSUE #227: [feature] Add support for decoding session state info in Ok packets

The `OkMessage.parse` currently does not attempt to decode session state informations, because it is not useful which at the time.

Currently, session state information of type `SESSION_TRACK_SYSTEM_VARIABLES` contains some potentially useful messages. e.g. `auto_commit`, `transaction_isolation`, or `redirect_url` (MariaDB only)

1. Make `Capability.SESSION_TRACK` available and disable it by default. Server will not respond session state info if it is disabled
2. Add session state information decoding in `OkMessage`, other types except `SESSION_TRACK_SYSTEM_VARIABLES` remain ignored
3. Add method `Map<String, String> getSystemVariables()` to `OkMessage` for getting value from the map of `SESSION_TRACK_SYSTEM_VARIABLES`

**Additional context**

See also [MySQL docs](https://dev.mysql.com/doc/dev/mysql-server/latest/page_protocol_basic_ok_packet.html#sect_protocol_basic_ok_packet_sessinfo) and [MariaDB docs](https://mariadb.com/kb/en/ok_packet/#session-state-info), [MariaDB `redirect_url` ticket](https://jira.mariadb.org/browse/MDEV-15935)

mirromutth:
Motivation:

Add support for `SESSION_TRACK_SYSTEM_VARIABLES` track info decoding

See also #227 

Modification:

- Make `Capability.SESSION_TRACK` available and disabled by default
- Add `SESSION_TRACK_SYSTEM_VARIABLES` decoding in `OkMessage`

Result:

Session state information decoding support

### DIFF
--- a/src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
+++ b/src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
@@ -21,8 +21,12 @@ import io.asyncer.r2dbc.mysql.ConnectionContext;
 import io.asyncer.r2dbc.mysql.constant.ServerStatuses;
 import io.asyncer.r2dbc.mysql.internal.util.VarIntUtils;
 import io.netty.buffer.ByteBuf;
+import org.jetbrains.annotations.Nullable;
 
 import java.nio.charset.Charset;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
 
 import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.requireNonNull;
 
@@ -33,6 +37,8 @@ import static io.asyncer.r2dbc.mysql.internal.util.AssertUtils.requireNonNull;
  * Note: OK message are also used to indicate EOF and EOF message are deprecated as of MySQL 5.7.5.
  */
 public final class OkMessage implements WarningMessage, ServerStatusMessage, CompleteMessage {
+
+    private static final int SESSION_TRACK_SYSTEM_VARIABLES = 0;
 
     private static final int MIN_SIZE = 7;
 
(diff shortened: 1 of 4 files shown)


src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
```diff
@@ -164,26 +182,79 @@ static OkMessage decode(boolean isEndOfRows, ByteBuf buf, ConnectionContext cont
...
+
+        switch (map.size()) {
+            case 0:
+                return Collections.emptyMap();
+            case 1: {
+                Map.Entry<String, String> entry = map.entrySet().iterator().next();
+                return Collections.singletonMap(entry.getKey(), entry.getValue());
+            }
+            default:
+                return map;
```
jchrys, line 259: Could we consider returning `Collections.unmodifiableMap(map)` to ensure the map remains immutable in all sizes?

jchrys (approved):
overall LGTM. 
just a nit

jchrys:
for personal reference) https://www.docs4dev.com/docs/en/mysql/5.7/reference/mysql-session-track-get-first.html

src/main/java/io/asyncer/r2dbc/mysql/message/server/OkMessage.java
mirromutth: I consider that this map should only have the `get` method called.

Maybe we can change `Map<String, String> getSystemVariables()` to `@Nullable String getSystemVariable(String key)`, then this map will never be changed outer than `OkMessage`, or?

jchrys: it makes sense. lgtm

jchrys (approved):

PR #231, opened 2024-02-13
Correct initialization of SSL tunnel tests

mirromutth:
Motivation:

If the server uses `caching_sha2_password`, the first time a client connects to the server, the server will require a native SSL connection. So all the SSL tunnel tests should be run after the `caching_sha2_password` initialization.

Steps to reproduce:

1. Restart a MySQL server 8.x, keep it authentication plugin as `caching_sha2_password`
2. Run `SslTunnelIntegrationTest` first, failed

```
Errors:
  SslTunnelIntegrationTest.sslTunnelConnectionTest » R2dbcPermissionDenied Access denied for user 'root'@'172.18.0.1' (using password: YES)
```

See also https://github.com/asyncer-io/r2dbc-mysql/actions/runs/7880628673/job/21502952115 . It skipped `MariaDbTextIntegrationTest` and `SslTunnelIntegrationTest` was the first case run, so it failed

Modification:

Add an initialization of `SslTunnelIntegrationTest`, it ensures the server has cached `root` user.

Result:

`SslTunnelIntegrationTest.sslTunnelConnectionTest` will not fail due to ordering of tests

### DIFF
--- a/src/test/java/io/asyncer/r2dbc/mysql/SslTunnelIntegrationTest.java
+++ b/src/test/java/io/asyncer/r2dbc/mysql/SslTunnelIntegrationTest.java
@@ -16,7 +16,6 @@
 
 package io.asyncer.r2dbc.mysql;
 
-
 import io.asyncer.r2dbc.mysql.constant.SslMode;
 import io.netty.bootstrap.Bootstrap;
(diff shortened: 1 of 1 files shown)


jchrys (approved):
Thanks a lot!

PR #232, opened 2024-02-13
Add LoopResources Configuration Support

### LINKED ISSUE #229: [feature] Support for Custom Loop Resources
Currently, r2dbc-mysql lacks support for custom loopResources. This limitation hinders the ability to optimize performance and manage resources efficiently, especially in scenarios requiring customized event loop management.

I propose the introduction of support for custom loop resources via `MySqlConnectionConfiguration` builder.

**Additional context**
The need for custom loop resource support is exemplified by the issue discussed in [r2dbc-pool#190](https://github.com/r2dbc/r2dbc-pool/issues/190), where it was observed that using custom loopResources could avoid accidental thread co-location, enhancing performance.

jchrys:
Motivation:
Enhances flexibility by allowing custom `LoopResources` configuration for optimized event loop management.

Modifications:
Introduced `LoopResources` configuration option in `MysqlConnectionConfiguration` and updated relevant documentation.

Result:
Enables performance tuning and flexibility, improving adaptability for various deployment scenarios. Resolves #229

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ ConnectionFactoryOptions options = ConnectionFactoryOptions.builder()
     .option(Option.valueOf("tcpKeepAlive"), true) // optional, default false
     .option(Option.valueOf("tcpNoDelay"), true) // optional, default false
     .option(Option.valueOf("compressionAlgorithms"), "zstd") // optional, default UNCOMPRESSED
+    .option(Option.valueOf("loopResources"), LoopResources.create("r2dbc")) // optional, default null, null means uses global tcp resources as loopResources (since 1.1.2)
     .option(Option.valueOf("autodetectExtensions"), false) // optional, default false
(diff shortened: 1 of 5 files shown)


mirromutth (approved):
LGTM

I have a question for this. We support `LOCAL INFILE` which is based on `AsynchronousFileChannel`. AFAIK, if we don't specify an `ExecutorService`, `AsynchronousFileChannel` will try to use the global `ThreadPool` by default. Should we apply the `LoopResources` to it?

I'm not sure because `LoopResources` are usually used as TCP resources, and `LOCAL INFILE` is not a frequently used feature.

jchrys:
I think we are okay to LoopResources for Tcp only. since AsynchronousFileChannel is using `min 0 max Integer.MAX_VALUE` executor by default so each opened will have a dedicated thread.(https://github.com/openjdk/jdk/blob/8765b176f97dbf334836f0aa6acd921d114304a9/src/java.base/share/classes/sun/nio/ch/SimpleAsynchronousFileChannelImpl.java#L46)

PR #238, opened 2024-02-16
Correct an unexpected fall through

mirromutth:
Motivation:

An unexpected fall through, which should have no impact. Because no packets start with `0xFE` during the command phase, the byte size is greater than 1, and not be OK or EOF.

Modification:

Correct an unexpected fall through in `ServerDecoder`.

Result:

Corrected an unexpected fall through.


jchrys (approved):
Wow Nice catch 👍

PR #240, opened 2024-02-19
Aligns timezone behavior with JDBC

### LINKED ISSUE #190: Aligning `serverZoneId` Behavior with `connectionTimeZone`(mysql-connector-j)

mirromutth:
Motivation:

Aligns timezone behavior with `mysql-connector-j`

See also #190 

Modification:

- Add `preserveInstants`, `connectionTimeZone` and `forceConnectionTimeZoneToSession` support
- Default `connectionTimeZone` to "LOCAL"
- Mark `serverZoneId` as deprecated. It will notice users to use `connectionTimeZone` instead
- Add `TimeZoneIntegrationTest` to test JDBC alignment of time zone behavior
- Correct `OffsetTimeCodec` to not convert time zone, `OffsetTime` is not an instant value

Result:

Users can now specify the local time zone to connection, allowing forcing the specified connection time zone on session.


jchrys:
Would it be possible to consider setting the default value for the connectionTimeZone property to 'LOCAL', aligning with the default used by mysql-connector-j? to enhance consistency for developers familiar with mysql-connector-j's defaults.

mirromutth:
> Would it be possible to consider setting the default value for the connectionTimeZone property to 'LOCAL', aligning with the default used by mysql-connector-j? to enhance consistency for developers familiar with mysql-connector-j's defaults.

This will break the current behavior. Currently, the configuration option `serverZoneId=null` would be equivalent to `connectionTimeZone=SERVER&preserveInstants=true&forceConnectionTimeZoneToSession=false`.

If we do, existing users (without `serverZoneId` configured) upgrade r2dbc-mysql to 1.1.2, it may unexpectedly cause date time types to use the local time zone. But, they are sent/responded to/from server, so they should be server time zone.

mirromutth:
In my personal opinion, I think we only need to write a migration guide wiki for new users who switch from jdbc to r2dbc.

It should contain:

- What attributes we have but jdbc does not
- What attributes we don’t have but jdbc does, why not, and how to migrate
- Which attributes we have with jdbc, but are different (such as naming and values), how to migrate them (**like time zone attributes**)

This also helps us clarify what we should do next.

WDYT? @jchrys

jchrys:
I 100% agree with your solution. I cannot deny that is the right way.
However, given the ongoing issue(a year long) with Micronaut, as highlighted here (https://github.com/micronaut-projects/micronaut-data/pull/2388), I face a challenge. Suggestions to set the default `connectionTimeZone` to `LOCAL` (as well as to match the behavior of mysql-connector-j) or to switch the default codec for `TIMESTAMP` and `DATETIME` fields to `LocalDateTime.class`(to aligned with mariadb-r2dbc and dev.miku's) were not accepted. (I think not accepting is reasonable)
I think we need another solution. But I don't have one right now for it.  I'm not trying to insist, but I would genuinely like to hear your thoughts on this. (including request changes on their end.) Could you please join the discussion and share your insights?

mirromutth:
@jchrys , the default `connectionTimeZone` is `LOCAL` now

The default type mappings of `DATETIME` will be fixed at other PR

jchrys:
@mirromutth  
I really appreciate for your efforts on this issue. Could you kindly elaborate on the rationale behind setting the default `connectionTimeZone` to `LOCAL` and adjusting `DATETIME` type mappings in the upcoming PR?

mirromutth:
> @mirromutth I really appreciate for your efforts on this issue. Could you kindly elaborate on the rationale behind setting the default `connectionTimeZone` to `LOCAL` and adjusting `DATETIME` type mappings in the upcoming PR?

@jchrys Based on the discussion in micronaut-projects/micronaut-data#2388, I realized that if we tried to use different time zone modes by default, it might cause more complicated problems for some existing projects.

Like I said, in my opinion, `ZonedDateTime` and server time zone makes more sense. However, if `LocalDateTime` and the JVM local time zone are widely accepted by other existing projects, we can't really change about it.

So, keep things simple.

jchrys:
> > @mirromutth I really appreciate for your efforts on this issue. Could you kindly elaborate on the rationale behind setting the default `connectionTimeZone` to `LOCAL` and adjusting `DATETIME` type mappings in the upcoming PR?
> 
> @jchrys Based on the discussion in [micronaut-projects/micronaut-data#2388](https://github.com/micronaut-projects/micronaut-data/pull/2388), I realized that if we tried to use different time zone modes by default, it might cause more complicated problems for some existing projects.
> 
> Like I said, in my opinion, `ZonedDateTime` and server time zone makes more sense. However, if `LocalDateTime` and the JVM local time zone are widely accepted by other existing projects, we can't really change about it.
> 
> So, keep things simple.

@mirromutth 
Thank you for your explanation. I got it.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConnectionContext.java
```diff
@@ -101,27 +109,33 @@ public CharCollation getClientCollation() {
...
+    @Override
+    public boolean isMariaDb() {
+        Capability capability = this.capability;
+        return (capability != null && capability.isMariaDb()) || serverVersion.isMariaDb();
     }
 
-    void setServerZoneId(ZoneId serverZoneId) {
-        if (this.serverZoneId != null) {
+    void setTimeZone(ZoneId timeZone) {
+        if (this.timeZone != null) {
```
jchrys: maybe `if (isTimeZoneInitialized())` ?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -283,7 +299,8 @@ public boolean equals(Object o) {
             tcpKeepAlive == that.tcpKeepAlive &&
             tcpNoDelay == that.tcpNoDelay &&
             Objects.equals(connectTimeout, that.connectTimeout) &&
-            Objects.equals(serverZoneId, that.serverZoneId) &&
+            Objects.equals(connectionTimeZone, that.connectionTimeZone) &&
```
jchrys, line 303: seems `preserveInstants` is missing

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -305,7 +322,8 @@ public boolean equals(Object o) {
     @Override
     public int hashCode() {
         return Objects.hash(isHost, domain, port, ssl, tcpKeepAlive, tcpNoDelay, connectTimeout,
-            serverZoneId, zeroDateOption, user, password, database, createDatabaseIfNotExist,
+            connectionTimeZone, forceConnectionTimeZoneToSession,
```
jchrys: seems `preserveInstants` is missing

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -316,7 +334,9 @@ public String toString() {
         if (isHost) {
             return "MySqlConnectionConfiguration{host='" + domain + "', port=" + port + ", ssl=" + ssl +
                 ", tcpNoDelay=" + tcpNoDelay + ", tcpKeepAlive=" + tcpKeepAlive +
-                ", connectTimeout=" + connectTimeout + ", serverZoneId=" + serverZoneId +
+                ", connectTimeout=" + connectTimeout +
```
jchrys, line 338: seems `preserveInstants` is missing

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionConfiguration.java
```diff
@@ -331,7 +351,9 @@ public String toString() {
         }
 
         return "MySqlConnectionConfiguration{unixSocket='" + domain +
-            "', connectTimeout=" + connectTimeout + ", serverZoneId=" + serverZoneId +
+            "', connectTimeout=" + connectTimeout +
```
jchrys, line 356: seems `preserveInstants` is missing

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/codec/LocalDateTimeCodec.java
```diff
@@ -29,6 +29,7 @@
 import java.time.LocalDate;
 import java.time.LocalDateTime;
 import java.time.LocalTime;
+import java.time.ZonedDateTime;
```
jchrys: unused imports?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/codec/ZonedDateTimeCodec.java
```diff
@@ -78,7 +79,9 @@ public boolean canDecode(MySqlColumnMetadata metadata, Class<?> target) {
     @Nullable
     private static ZonedDateTime decode0(ByteBuf value, boolean binary, CodecContext context) {
         LocalDateTime origin = LocalDateTimeCodec.decodeOrigin(value, binary, context);
-        return origin == null ? null : ZonedDateTime.of(origin, context.getServerZoneId());
+        ZoneId zoneId = context.isPreserveInstants() ? context.getTimeZone() : ZoneId.systemDefault();
+
+        return origin == null ? null : ZonedDateTime.of(origin, zoneId);
```
jchrys: if origin is null, I think we can remove unnecessary calculation

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/internal/util/StringUtils.java
```diff
@@ -70,6 +79,38 @@ public static String extendReturning(String sql, String returning) {
...
+     * <p>
+     * Note: since java 14.0.2, 11.0.8, 8u261 and 7u271, America/Nuuk is already renamed from America/Godthab.
+     * See also <a href="https://mm.icann.org/pipermail/tz-announce/2020-April/000058.html">tzdata2020a</a>
+     *
+     * @param zoneId the time zone string
+     * @return the normalized {@link ZoneId}
+     * @throws java.time.DateTimeException if the time zone string has an invalid format
+     * @throws java.time.zone.ZoneRulesException if the time zone string cannot be found
+     */
+    public static ZoneId parseZoneId(String zoneId) {
```
jchrys, line 94: could we add requireNonNull since this is an utility class

r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/TimeZoneIntegrationTest.java
```diff
@@ -0,0 +1,360 @@
...
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+import static java.util.Objects.requireNonNull;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.within;
+
+/**
+ * Integration tests for aligning time zone configuration options with jdbc.
+ */
```
jchrys, line 45: will `@Isolated` is required? since it changes default timezone temporarily. it may affect other tests when tests are executed concurrently.

jchrys (commented):
overall lgtm.
just a nit and suggestions.

jchrys (approved):
LGTM.
Given this changes default behavior(?), which version do you think best suits this change? (wonder if 1.1.2 would be enough for this changes WDYT?)

mirromutth:
I think we can release it and #243 on 1.1.2. This version mainly deals with date time and time zone problems.

While this may bring some breaking changes, we can add some wiki for it.

jchrys:
> I think we can release it and #243 on 1.1.2. This version mainly deals with date time and time zone problems.
> 
> While this may bring some breaking changes, we can add some wiki for it.

Got it. Thanks :D

PR #243, opened 2024-02-21
Default to use LocalDateTime for DATETIME

### LINKED ISSUE #166: Discussion on Default Codecs for R2DBC MySQL

mirromutth:
Motivation:

See also #166 

Modification:

Change to use default `LocalDateTime` for `DATETIME`

Result:

Default to use `LocalDateTime` for `DATETIME`

### DIFF
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/constant/MySqlType.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/constant/MySqlType.java
@@ -23,6 +23,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.time.LocalTime;
 import java.time.ZonedDateTime;
 
@@ -271,7 +272,7 @@ public enum MySqlType implements Type {
     /**
      * A date time type. It does not contain timezone. It uses string format to transfer the value.
      */
-    DATETIME(MySqlType.ID_DATETIME, ZonedDateTime.class),
+    DATETIME(MySqlType.ID_DATETIME, LocalDateTime.class),
 
     /**
      * A year type. It contains neither leap year information nor timezone.


jchrys:
Looks like merging PR #240 fixes the Micronaut issue. Could you please kindly elaborate on why we need this additional PR? (because it seems half revert of previous change (https://github.com/asyncer-io/r2dbc-mysql/blob/bf4874951e440c8c7ae66f52650918993966d2ad/src/main/java/dev/miku/r2dbc/mysql/MySqlColumnMetadata.java#L135C3-L137C44))

mirromutth:
> Looks like merging PR #240 fixes the Micronaut issue. Could you please kindly elaborate on why we need this additional PR? (because it seems half revert of previous change (https://github.com/asyncer-io/r2dbc-mysql/blob/bf4874951e440c8c7ae66f52650918993966d2ad/src/main/java/dev/miku/r2dbc/mysql/MySqlColumnMetadata.java#L135C3-L137C44))

It just a very simple idea, which may not be correct: Since there are some problems caused by non-aligned processing of date times and time zones, so I'm trying to ensure that the behavior as same as JDBC.

In JDBC, `DATETIME` would map to `LocalDateTime` and `TIMESTAMP` to `Timestamp` which should be re-map to `ZonedDateTime`.

jchrys:
> > Looks like merging PR #240 fixes the Micronaut issue. Could you please kindly elaborate on why we need this additional PR? (because it seems half revert of previous change (https://github.com/asyncer-io/r2dbc-mysql/blob/bf4874951e440c8c7ae66f52650918993966d2ad/src/main/java/dev/miku/r2dbc/mysql/MySqlColumnMetadata.java#L135C3-L137C44))
> 
> @jchrys It just a very simple idea, which may not be correct: Since there are some problems caused by non-aligned processing of date times and time zones, so I'm trying to ensure that the behavior as same as JDBC.
> 
> In JDBC, `DATETIME` would map to `LocalDateTime` and `TIMESTAMP` to `ZonedDateTime`.

Thanks. I understand now, it's clear.

jchrys (approved):
lgtm

PR #248, opened 2024-02-27
Correct version auto-update in workflow

mirromutth:
Motivation:

The version of `test-native-image` will not be updated after the release is completed, which causes CI to fail after releasing.

Modification:

- Upgrade `test-native-image` to `1.1.3-SNAPSHOT`
- Correct version auto-update in workflow

Result:

Now the version of `test-native-image` should be updated automatically.


jchrys (approved):
lgtm

PR #252, opened 2024-02-29
Correct transaction statement building

mirromutth:
Motivation:

Currently, `START TRANSACTION` does not allow `READ ONLY` and `WITH CONSISTENT SNAPSHOT` to be used at the same time. Multiple transaction properties need to be separated by commas. e.g. `START TRANSACTION WITH CONSISTENT SNAPSHOT, READ ONLY`

Modification:

- Correct transaction statement building
- Correct incorrect description in javadoc of `MySqlTransactionDefinition`
- Add unit test for transaction statement building
- Add integration test for start transaction with consistent and read only

Result:

`READ ONLY` and `WITH CONSISTENT SNAPSHOT` can be used simultaneously in a transaction.

### DIFF
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
@@ -1281,20 +1281,28 @@ final class StartTransactionState extends AbstractTransactionState {
         return false;
     }
 
-    private static String buildStartTransaction(TransactionDefinition definition) {
+    /**
+     * Visible for testing.
+     *
+     * @param definition the transaction definition
+     * @return the {@code START TRANSACTION} statement
+     */
+    static String buildStartTransaction(TransactionDefinition definition) {
         Boolean readOnly = definition.getAttribute(TransactionDefinition.READ_ONLY);
         Boolean snapshot = definition.getAttribute(MySqlTransactionDefinition.WITH_CONSISTENT_SNAPSHOT);
 
-        if (readOnly == null && (snapshot == null || !snapshot)) {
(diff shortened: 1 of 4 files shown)


r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
```diff
@@ -1311,12 +1319,12 @@ private static String buildStartTransaction(TransactionDefinition definition) {
...
-            if (readOnly) {
-                builder.append(" READ ONLY");
-            } else {
-                builder.append(" READ WRITE");
+        if (Boolean.TRUE.equals(readOnly)) {
+            if (!first) {
+                builder.append(',');
             }
+
+            builder.append(" READ ONLY");
```
jchrys: (if not intended)I believe we may need to include the `READ WRITE` unless it is explicitly set to read-only.
without `READ WRITE`, it will fallback to the default mode, which differs from how it behaved previously.

jchrys (changes requested):
Overall, LGTM, but I do have one question.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/QueryFlow.java
mirromutth: You are right, my mistake.

See [r2dbc-mariadb](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/blob/master/src/main/java/org/mariadb/r2dbc/client/SimpleClient.java#L405), I'm writing something similar to it.

While it's not a big deal, considering that different distributions **may** have inconsistent implementations of `SET TRANSACTION READ ONLY/WRITE` , it might be better for us to continue using `READ WRITE` explicitly.

jchrys (approved):
lgtm

PR #253, opened 2024-02-29
Extract API to api package

### LINKED ISSUE #251: [feature] Extract public API into API package

Now we are exposing custom API extensions such as `MySqlBatch`, `MySqlStatement`, etc. So I think we can expose that API using a stable design.

Extract API will help us:

1. Standardize our interface
2. Let us clearly know which implementations are not consistent with r2dbc spec, like #244
3. Issue #245 can be resolved in an easier way
4. [Java 9 Modules](https://dev.java/learn/modules/) support can be added in an easier way
5. Make it easier for users or new contributors to understand the source code

Extract these `public interface`s to `api` package:

- `MySqlBatch`
- `MySqlConnection`
- `MySqlConnectionMetadata`
- `MySqlTransactionDefinition`
- `MySqlTypeMetadata`
- `MySqlStatement`
- `MySqlReadable`
- `MySqlOutParameters`
- `MySqlRow`
- `MySqlRowMetadata`
- `MySqlOutParametersMetadata`
- `MySqlReadableMetadata`
- `MySqlColumnMetadata`
- `MySqlOutParameterMetadata`
- `MySqlResult`
   - `MySqlResult.MySqlSegment` or just `MySqlResult.Segment`
   - `MySqlResult.MySqlRowSegment` or just `MySqlResult.RowSegment`
   - `MySqlResult.MySqlOutSegment` or just `MySqlResult.OutSegment`
   - `MySqlResult.MySqlUpdateCount` or just `MySqlResult.UpdateCount`
   - `MySqlResult.MySqlMessage` or just `MySqlResult.Message`

**Additional context**

The `extension.Extension` and `extension.CodecRegistrar` are our extension SPI instead of an API, so keep as is.

See also [`r2dbc-mariadb`](https://github.com/mariadb-corporation/mariadb-connector-r2dbc/tree/master/src/main/java/org/mariadb/r2dbc/api) and [`r2dbc-postgres`](https://github.com/pgjdbc/r2dbc-postgresql/tree/main/src/main/java/io/r2dbc/postgresql/api) does.

mirromutth:
Motivation:

Extract public API to api package, see also #251 

Modification:

Most of changes are to **use strongly typed interfaces** in the new `api` package instead of using `r2dbc-spi` interface directly. So, the modification seems large, but it is mainly changes such as `implements ColumnMetadata` -> `implements MySqlColumnMetadata`.

- Extract API to `api` package
   - **Use strongly typed interfaces**
   - Change `Codec`/`Codecs` to use `ReadableMetadata` (`MySqlReadableMetadata`) instead of `ColumnMetadata`
   - Change the access modifier of non-public APIs to make them as private as possible, e.g. `public class NonPublic` -> `[package-private] class NonPublic`
   - Move `MySqlStatement` to `api.MySqlStatement`
   - Write more MySQL-specific API details in javadoc instead of inherited docs
   - Merge `ColumnDefinition` and `MySqlTypeMetadata` to new `MySqlTypeMetadata`, which is implemented by `api.MySqlNativeTypeMetadata`
- Mark old `MySqlTransactionDefinition` as deprecated, use `api.MySqlTransactionDefinition` instead
- Mark `ConsistentSnapshotEngine` as deprecated, use `String` instead of an `enum`
   - It seems like an over-engineered and hard to expand. e.g. A user wants to use an engine other than RocksDB or InnoDB, then the user needs to wait for us to add the element to the `enum`
   - This is a distribution-specific feature, we have no reason to maintain an `enum` for this
- Find some bugs and add FIXME, they should be fixed in other tickets

Result:

Standardized our public interfaces


r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConsistentSnapshotEngine.java
```diff
@@ -19,7 +19,12 @@
 /**
  * The engine of {@code START TRANSACTION WITH CONSISTENT [engine] SNAPSHOT} for Facebook/MySQL or similar
  * syntax.
+ *
+ * @deprecated use directly {@link String} instead, e.g. {@code "ROCKSDB"}
```
jchrys: In the Javadoc deprecated section, should we include "since ${version}" as is customary with others?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlDataRow.java
```diff
@@ -69,33 +69,15 @@ public <T> T get(String name, Class<T> type) {
...
-    /**
-     * Returns the value for a column in this row. The value can be a parameterized type.
-     *
-     * @param index the index of the column starting at {@code 0}.
-     * @param type  the parameterized type of item to return.
-     * @param <T>   the type of the item being returned.
-     * @return the value for a column in this row. Value can be {@code null}.
-     * @throws IllegalArgumentException if {@code name} or {@code type} is {@code null}.
-     */
-    @Nullable
```
jchrys, line 81: Might `@Nullable` be missing?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlDataRow.java
```diff
@@ -69,33 +69,15 @@ public <T> T get(String name, Class<T> type) {
...
-    /**
-     * Returns the value for a column in this row. The value can be a parameterized type.
-     *
-     * @param name the name of the column.
-     * @param type the parameterized type of item to return.
-     * @param <T>  the type of the item being returned.
-     * @return the value for a column in this row. Value can be {@code null}.
-     * @throws IllegalArgumentException if {@code name} or {@code type} is {@code null}.
-     */
-    @Nullable
```
jchrys, line 98: Might `@Nullable` be missing?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadata.java
```diff
@@ -59,16 +117,20 @@ public boolean equals(Object o) {
...
-        return 31 * id + definition.hashCode();
+        int result = 31 * typeId + (int) definitions;
+        return 31 * result + collationId;
     }
 
     @Override
     public String toString() {
-        return "MySqlTypeMetadata(" + id + ", " + definition + ')';
+        return "MySqlTypeMetadata{typeId=" + typeId +
+            ", definitions=0x" + Integer.toHexString(definitions & ALL_USED) +
```
jchrys: seems `& ALL USED` is applied when created

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ParametrizedStatementSupport.java
jchrys, line 1: might be a typo of `Parameterized`?

r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadataTest.java
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright 2023 asyncer.io projects
```
jchrys: ` * Copyright 2024 asyncer.io projects`

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/api/SimpleTransactionDefinition.java
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    @Override
+    public int hashCode() {
+        return options.hashCode();
+    }
+
+    @Override
+    public String toString() {
+        return "MySqlTransactionDefinition" + options;
```
jchrys: `SimpleTransactionDefinition` ?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlTransactionDefinition.java
```diff
@@ -33,7 +33,9 @@
  * and 1073741824.
  *
  * @since 0.9.0
+ * @deprecated use {@link io.asyncer.r2dbc.mysql.api.MySqlTransactionDefinition} instead.
```
jchrys: seems deprecated since is missing

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlDataRow.java
mirromutth: `Override` will make it inherit `Nullable` from `api.MySqlRow`

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ParametrizedStatementSupport.java
mirromutth: The `Parameterized` seems to be a legacy typo, and if all of this typo was to be resolved, it would involve a lot of code and javadoc in itself. I prefer to resolve it in another PR if necessary.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlDataRow.java
jchrys: Oh, got it. Thanks!

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ParametrizedStatementSupport.java
jchrys: Sure. make sense :D

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConsistentSnapshotEngine.java
mirromutth: Yep, before I committed I have no idea about which version it was supposed to be released, it seems good to fill in 1.1.3 now

jchrys: Oh, You're right. since this is not backward compatible(possible breaking compatibility). I think we need to increment major or minor. WDYT?

jchrys (approved):
lgtm

mirromutth:
> Oh, You're right. since this is not backward compatible(possible breaking compatibility). I think we need to increment major or minor. WDYT?

@jchrys

Those users who will not be affected by this modification:

- Use this driver with spi `ConnectionFactories`
- Only used `MySqlConnectionConfiguration` to configure somethings, and did not involve `CodecRegistrar`

This modification will cause a compilation warning for those users:

- User is using `MySqlTransactionDefinition` and `ConstantSnapshotEngine` to use release-specific transaction properties. They should see `@Deprecated`

This modification will cause those users who need to modify the code to:

- User is using `CodecRegistrar` to extend custom `Codec`. They should use `MySqlReadableMetadata` instead of the old `MySqlColumnMetadata` as any `Codec` should support `OutParameters`, which is a necessary change for the purpose of complying with the r2dbc spec
- User is using old `MySqlConnection`, `MySqlRow`, `MySqlStatement`, etc. Their code may need to be changed to `api.MySqlXxx`, which makes sense because it was unclear which parts of the original API were public and which parts were not. Now with the `api` package, this will not be the case

I have 3 suggestions:

- Consider a cumulative update to 1.2.0, which will provide clearly needed new features and fix all currently known issues. This should require more work
- Just use 1.1.3, it is the first version for public API, and `CodecRegitrar` changes is necessary for supporting `OUT` parameters in r2dbc spec
- Consider skipping some patch versions, e.g. 1.1.10 is next

jchrys:
> > Oh, You're right. since this is not backward compatible(possible breaking compatibility). I think we need to increment major or minor. WDYT?
> 
> @jchrys
> 
> Those users who will not be affected by this modification:
> 
> * Use this driver with spi `ConnectionFactories`
> * Only used `MySqlConnectionConfiguration` to configure somethings, and did not involve `CodecRegistrar`
> 
> This modification will cause a compilation warning for those users:
> 
> * User is using `MySqlTransactionDefinition` and `ConstantSnapshotEngine` to use release-specific transaction properties. They should see `@Deprecated`
> 
> This modification will cause those users who need to modify the code to:
> 
> * User is using `CodecRegistrar` to extend custom `Codec`. They should use `MySqlReadableMetadata` instead of the old `MySqlColumnMetadata` as any `Codec` should support `OutParameters`, which is a necessary change for the purpose of complying with the r2dbc spec
> * User is using old `MySqlConnection`, `MySqlRow`, `MySqlStatement`, etc. Their code may need to be changed to `api.MySqlXxx`, which makes sense because it was unclear which parts of the original API were public and which parts were not. Now with the `api` package, this will not be the case
> 
> I have 3 suggestions:
> 
> * Consider a cumulative update to 1.2.0, which will provide clearly needed new features and fix all currently known issues. This should require more work
> * Just use 1.1.3, it is the first version for public API, and `CodecRegitrar` changes is necessary for supporting `OUT` parameters in r2dbc spec
> * Consider skipping some patch versions, e.g. 1.1.10 is next

@mirromutth
It appears that we have yet to adopt a unified approach to versioning. It would be advantageous for us to agree on a versioning strategy. For example, whether to follow Strict SemVer (https://semver.org/) or a more relaxed version of SemVer, etc. In my opinion, I would like to suggest a slightly relaxed form of SemVer. For backward incompatible changes or new features, we could update the minor version. For backward compatible changes or new features, we could update the patch version. And for major versions, we could reserve those for significant or 'kaboom' changes. In this case, I would like to suggest updating to version 1.2.0.

mirromutth:
@jchrys Thanks.

I consider avoiding the version growing too fast, because actually this PR did not introduce much new content.

Maybe we should delay merging this PR until current known bugs are resolved. Then we consider upgrading to 1.2? WDYT?

jchrys:
@mirromutth 
Thanks for sharing your thought.
In that case, proceeding with the originally planned version 1.1.3 seems appropriate.

Could you please confirm if the following version plan aligns with what you have in mind? if right we can follow that rule.

Major: Increment for big bang changes.
Minor: Increment when resolved all known bugs or for relatively big changes.
Patch: increment for everything else.

mirromutth:
LGTM, the major version increment of `r2dbc-spi` is also a big change

jchrys:
> LGTM, the major version increment of `r2dbc-spi` is also a big change

Got it. I agree.

jchrys (approved):

PR #257, opened 2024-03-06
Revise `RowMetaData#getColumnMetadata(String)`

### LINKED ISSUE #244: [bug]align with`RowMetaData#getColumnMetadata(String)` specification
**Describe the bug**
The specification states 
```
@param name the name of the column.  
Column names are case-insensitive.  
When a get method contains several columns with same name, 
then the value of the >> first matching column << will be returned
```
However, the current behavior does not align with the specification.

**To Reproduce**
```Java
connection.createStatement(SELECT 1 as t, 2 as t, 3 as t).execute()
                .flatMap(result -> Mono.from(result.map((row, metadata) -> row.get("t", Integer.class))))
                .doOnNext(System.out::println);
```
Prints `2`

**Expected behavior**
It should print `1`

**Screenshots**

**Additional context**

jchrys:
Motivation:
The current behavior does not align with the specification

Modifications:
implement the intended behavior as per the specification.

Result:
Correct behavior


r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/MySqlRowDescriptorTest.java
```diff
@@ -0,0 +1,65 @@
...
+        MySqlColumnDescriptor[] metadata = new MySqlColumnDescriptor[names.length];
+        for (int i = 0; i < names.length; ++i) {
+            metadata[i] =
+                    new MySqlColumnDescriptor(i, (short) 0, names[i], 0, 0, 0, 1);
+        }
+        return new MySqlRowDescriptor(metadata);
+    }
+
+
+}
```
mirromutth: Maybe a line break?

mirromutth (approved):
LTGM, only a code style nit

r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/MySqlRowDescriptorTest.java
jchrys: Sure, Let me add it.
(check style should've thrown an error. let me check)

jchrys: Currently, the Checkstyle plugin does not analyze source code located under the test directory. For the time being, I prefer to keep it.

Issue #228, opened 2024-02-09
Add support for MariaDB connection redirection


mirromutth:
In the Ok packet of MariaDB, the session state information may contain `redirect_url` and the driver can keep the old connection, or redirect to the new URL.

> Supporting redirection mechanism is helpful for proxy-based scenario. It is more like HTTP redirects or Oracle redirected connections. Clients get the servers’ address from proxies and connect to servers transparently without latency and computing resource wasted.

It is based on a common feature of MySQL and MariaDB, #227

See also https://jira.mariadb.org/browse/MDEV-15935 and https://mariadb.com/kb/en/ok_packet/#session_track_system_variables .

Issue #89, opened 2023-04-13
[feature]Supports Multiple host configuration.


tabdelbari:
**Is your feature request related to a problem? Please describe.**
The probleme is that tve driver doesn't support multiple host configuration 

**Describe the solution you'd like**
support multiple hosts configuration, for example: r2dbc:mysql//host1,host2/db

jchrys:
Hello, @tabdelbari.
Thank you for bringing this feature request to our attention. I'll be taking up the task to work on implementing multiple host configurations. 

Considering that the R2DBC SPI already supports multiple hosts, it would be beneficial for our project to have this feature as well.

Similar requests have been made in the past, and there's a clear demand for this feature
- https://github.com/mirromutth/r2dbc-mysql/issues/169
- https://github.com/mirromutth/r2dbc-mysql/issues/183

jchrys:
reference
- https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-multi-host-connections.html
- https://r2dbc.io/spec/1.0.0.RELEASE/spec/html/#overview.connection.url
-

tabdelbari:
Thank you so much @jchrys

mirromutth:
It requires following tasks:

1. Connection strategy when creating the connection
   - https://dev.mysql.com/doc/connector-j/en/connector-j-multi-host-connections.html
   - https://mariadb.com/kb/en/failover-and-high-availability-with-mariadb-connector-j
2. DNS SRV Records support: https://dev.mysql.com/doc/refman/8.0/en/connecting-using-dns-srv.html
3. Reconnection strategy: `autoReconnect`/`maxReconnects`/`reconnectAtTxEnd`/`initialTimeout` option. It should be reconnected only if the connection is in auto-commit mode (not in a transaction).
4. If `secondaryBeforeRetrySource` or `requestsBeforeRetrySource` is set, the **default** (no name) high availability mode will attempt to return the first host/address even if no communication exception occurs
5. `retriesAllDown` option or alternative. If the connection fails for all hosts/addresses, whether to retry them all

These we should not consider:

- Replication mode: `io.r2dbc.spi.Connection` cannot be set to a read-only connection, so this mode will always use the first host/address
- `failOverReadOnly` option: same reason as above
- `autoReconnectForPools` option: force reconnect even the connection is in a transaction or `autoCommit=false`. Not sure it is necessary
[...]

PR #260, opened 2024-03-14
Reduce retained references to ConnectionContext

mirromutth:
Motivation:

A small refactoring of `ConnectionContext` usage.

It is based on challenges #228 and #89. If a MySQL connection is reconnected or redirected, the `ConnectionContext` needs to be recreated. If we accidentally retain a reference to `ConnectionContext` outside the `client` package, it may cause some unexpected behavior.

Of course, there are some counterexamples here, such as `MySqlDataRow` that needs to retain the `CodecContext` in which it was responded to. Regardless of whether the connection changes, the `CodecContext` should be always the context of the connection which was responding to this row, instead of the current connection.

Modification:

- Reduce publicity
- Prepare to support redirect and reconnect

Result:

Remove `ConnectionContext` reference that's retained outside `client` package, except those required by codecs


jchrys (approved):
LGTM

PR #268, opened 2024-03-28
Add support for SQL mode `NO_BACKSLASH_ESCAPES`

### LINKED ISSUE #267: [feature] Support sql mode `NO_BACKSLASH_ESCAPES`

When `sql_mode` contains `NO_BACKSLASH_ESCAPES`, MySQL server does not allow the use of `\` as an escape character. In this mode, we should not use `\` to escape strings.

For example:

```sql
CREATE TABLE escape_test(id INT NOT NULL AUTO_INCREMENT PRIMARY KEY, value VARCHAR(50) NOT NULL);

INSERT INTO escape_test (`value`) VALUES ('my\ndata');

SELECT * FROM escape_test WHERE value = 'my
data'; -- 1 row
SELECT * FROM escape_test WHERE value = 'my\ndata'; -- 1 row

SET SESSION sql_mode = 'NO_BACKSLASH_ESCAPES';

SELECT * FROM escape_test WHERE value = 'my
data'; -- 1 row
SELECT * FROM escape_test WHERE value = 'my\ndata'; -- 0 row
```

- Make `ServerStatuses.NO_BACKSLASH_ESCAPES` available
- If `NO_BACKSLASH_ESCAPES` is set, disable escape
- For the `innodb_lock_wait_timeout`, use `SHOW VARIABLES LIKE 'innodb\_lock\_wait\_timeout'` instead of `SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'`

mirromutth:
Motivation:

Resolves #267 

Modification:

- `ParamWriter` skips backslash escaping if `NO_BACKSLASH_ESCAPES` is set
- Load `innodb_lock_wait_timeout` without escaping

And unit tests and integration tests for `NO_BACKSLASH_ESCAPES`.

Result:

SQL mode `NO_BACKSLASH_ESCAPES` supported

### DIFF
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConnectionContext.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConnectionContext.java
@@ -216,6 +216,10 @@ public final class ConnectionContext implements CodecContext {
         return (capability != null && capability.isMariaDb()) || serverVersion.isMariaDb();
     }
 
+    public boolean isNoBackslashEscapes() {
+        return (serverStatuses & ServerStatuses.NO_BACKSLASH_ESCAPES) != 0;
+    }
+
     @Override
     public ZeroDateOption getZeroDateOption() {
         return zeroDateOption;
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/InitFlow.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/InitFlow.java
@@ -209,31 +209,31 @@ final class InitFlow {
         Codecs codecs,
         @Nullable Duration lockWaitTimeout
     ) {
-        return new TextSimpleStatement(client, codecs, "SHOW VARIABLES LIKE 'innodb\\\\_lock\\\\_wait\\\\_timeout'")
-            .execute()
-            .flatMap(r -> r.map(readable -> {
-                String value = readable.get(1, String.class);
+        return new TextSimpleStatement(
+            client,
+            codecs,
+            "SHOW VARIABLES LIKE 'innodb_lock_wait_timeout'"
+        ).execute().flatMap(r -> r.map(readable -> {
+            String value = readable.get(1, String.class);
 
-                if (value == null || value.isEmpty()) {
-                    return data;
-                } else {
-                    return data.lockWaitTimeout(Duration.ofSeconds(Long.parseLong(value)));
+            if (value == null || value.isEmpty()) {
+                return data;
+            } else {
+                return data.lockWaitTimeout(Duration.ofSeconds(Long.parseLong(value)));
+            }
+        })).single(data).flatMap(d -> {
+            if (lockWaitTimeout != null) {
+                // Do not use context.isLockWaitTimeoutSupported() here, because its session variable is not set
+                if (d.lockWaitTimeoutSupported) {
+                    return QueryFlow.executeVoid(client, StringUtils.lockWaitTimeoutStatement(lockWaitTimeout))
+                        .then(Mono.fromSupplier(() -> d.lockWaitTimeout(lockWaitTimeout)));
                 }
-            }))
-            .single(data)
-            .flatMap(d -> {
-                if (lockWaitTimeout != null) {
-                    // Do not use context.isLockWaitTimeoutSupported() here, because its session variable is not set
-                    if (d.lockWaitTimeoutSupported) {
-                        return QueryFlow.executeVoid(client, StringUtils.lockWaitTimeoutStatement(lockWaitTimeout))
-                            .then(Mono.fromSupplier(() -> d.lockWaitTimeout(lockWaitTimeout)));
-                    }
 
-                    logger.warn("Lock wait timeout is not supported by server, ignore initial setting");
-                    return Mono.just(d);
-                }
+                logger.warn("Lock wait timeout is not supported by server, ignore initial setting");
                 return Mono.just(d);
-            });
+            }
+            return Mono.just(d);
+        });
     }
 
     private static Mono<SessionState> loadSessionVariables(Client client, Codecs codecs) {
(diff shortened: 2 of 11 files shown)


r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlSimpleConnection.java
```diff
@@ -563,7 +563,7 @@ private static Mono<SessionData> loadSessionVariables(Client client, Codecs code
     }
 
     private static Mono<SessionData> loadInnoDbEngineStatus(SessionData data, Client client, Codecs codecs) {
-        return new TextSimpleStatement(client, codecs, "SHOW VARIABLES LIKE 'innodb\\\\_lock\\\\_wait\\\\_timeout'")
+        return new TextSimpleStatement(client, codecs, "SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'")
```
jchrys: it seems this query won't work if `isNoBackslashEscapes() == true` :(

jchrys (changes requested):

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlSimpleConnection.java
mirromutth: I tested, and it should work on `mysql:5.5`, `mariadb:latest` <del>and `mysql:latest`</del>.

See the following:

```sql
SET sql_mode = 'STRICT_TRANS_TABLES';

SHOW VARIABLES LIKE 'innodb\_lock\_wait\_timeout'; -- 1 row
SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'; -- 1 row

SET sql_mode = 'STRICT_TRANS_TABLES,NO_BACKSLASH_ESCAPES';

SHOW VARIABLES LIKE 'innodb\_lock\_wait\_timeout'; -- 1 row
SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'; -- 0 row
```

If `NO_BACKSLASH_ESCAPES` is set, `\_` will be equivalent to `\\_` on `NO_BACKSLASH_ESCAPES` is unset. The backslash is used by `LIKE` matching instead of escaping.

If `NO_BACKSLASH_ESCAPES` is not set, the `\_` escape will be ignored, and the original string `\_` will be obtained by matching.

Therefore, regardless of whether `NO_BACKSLASH_ESCAPES` is set, the `LIKE` match will obtain the `\_` pattern and try to strictly match the `_` character.

jchrys: ~Could you provide the version of your MySQL client and details about your environment? I received a different result.. I used MySQL 8.3.0 Community edition(mysql:latest docker tag) and mysql client (`mysql  Ver 8.2.0 for macos14.0 on arm64 (Homebrew)`).~

```
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 9
Server version: 8.3.0 MySQL Community Server - GPL

Copyright (c) 2000, 2023, Oracle and/or its affiliates.

Oracle is a registered trademark of Oracle Corporation and/or its
affiliates. Other names may be trademarks of their respective
owners.

Type 'help;' or '\h' for help. Type '\c' to clear the current input statement.

mysql> SET sql_mode = 'STRICT_TRANS_TABLES';
Query OK, 0 rows affected, 1 warning (0.00 sec)

mysql> 
mysql> SHOW VARIABLES LIKE 'innodb\_lock\_wait\_timeout'; -- 1 row
+--------------------------+-------+
| Variable_name            | Value |
+--------------------------+-------+
| innodb_lock_wait_timeout | 50    |
+--------------------------+-------+
1 row in set (0.01 sec)

mysql> SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'; -- 1 row
+--------------------------+-------+
| Variable_name            | Value |
+--------------------------+-------+
| innodb_lock_wait_timeout | 50    |
+--------------------------+-------+
1 row in set (0.01 sec)

mysql> 
mysql> SET sql_mode = 'STRICT_TRANS_TABLES,NO_BACKSLASH_ESCAPES';
Query OK, 0 rows affected, 1 warning (0.00 sec)

mysql> 
mysql> SHOW VARIABLES LIKE 'innodb\_lock\_wait\_timeout'; -- 1 row
Empty set (0.00 sec)

mysql> SHOW VARIABLES LIKE 'innodb\\_lock\\_wait\\_timeout'; -- 0 row
Empty set (0.01 sec)
```

jchrys (approved):

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlSimpleConnection.java
jchrys: @mirromutth 
Understood. Let's proceed according to the specifications. It seems that My local environment is acting strangely.(https://dev.mysql.com/doc/refman/8.0/en/string-comparison-functions.html#operator_like)

mirromutth: @jchrys You're right, it's my fault. For some reason in other projects, I have a 5.7-compatible internal distribution marked as `mysql:latest` on my local environment.

I tried again in a clean environment and found that in MySQL 8.0 and later, this behavior is broken. MySQL 5.7 and MariaDB 11.2.2 can match this pattern, MySQL 8.0 and above cannot.

mirromutth: It is changed to `SHOW VARIABLES LIKE 'innodb_lock_wait_timeout'`

jchrys (approved):
LGTM 😄

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlSimpleConnection.java
jchrys: may need to report this issue D

PR #286, opened 2024-08-03
Handle `BIT(1)` as `Boolean` by Default

### LINKED ISSUE #277: [feature] return a scalar value instead of a buffer for bit fields, such as Byte.
r2dbc has ConverterNotFoundException [java.nio.HeapByteBuffer] to type [java.lang.Boolean] on boot 3.3.0
https://github.com/spring-projects/spring-data-relational/issues/1824

return a scalar value instead of a buffer for bit fields, such as Byte.

**Additional context**
https://github.com/spring-projects/spring-data-relational/issues/1824#issuecomment-2185738223

jchrys:
Motivation:
BIT(1) should be treated as `Boolean`.

Modification:
set `Boolean.class` as `BIT(1)'s default java type.

Result:
Ensures `BIT(1)` is handled as `Boolean` by default. like mysql-connector-j's implementation. 
Resolves #277


jchrys:
type conversion chart to ref: https://dev.mysql.com/doc/connector-j/en/connector-j-reference-type-conversions.html

Issue #254, opened 2024-03-01
Support for MariaDB Extend Type Info


mirromutth:
If `Capability.MARIADB_CLIENT_EXTENDED_TYPE_INFO` is set, the [Column Definition Packet](https://mariadb.com/kb/en/result-set-packets/#column-definition-packet) will contain an [extend type info](https://mariadb.com/kb/en/result-set-packets/#extended-type-info)

MariaDB returns `JSON` data as `VARCHAR` (`MYSQL_TYPE_STRING`), so reading the extend type info is useful to identify the correct `MySqlType`.

PR #288, opened 2024-09-06
Added support for MariaDb extend type info

svats0001:
Hi @mirromutth ,
#254 

Motivation:
If Capability.MARIADB_CLIENT_EXTENDED_TYPE_INFO is set, the Column Defintion Packet will contain extended type information. MariaDb returns JSON data as VARCHAR (MYSQL_TYPE_STRING) so reading the extended type information can be useful to identify the correct MySqlType.

Modification:
Capability: Uncommented MARIADB_CLIENT_EXTENDED_TYPE_INFO and added method that checks if this capability is enabled in bitmap.
MySqlColumnDescriptor: Altered constructor parameter to include nullable extendedTypeInfo field that is used to create MySqlTypeMetadata.
MySqlTypeMetadata: Added nullable extendedTypeInfo field and added isMariaDbJson method to check if the value of extendedTypeInfo equals 'json'. Included extendedTypeInfo in equals, hashcode and toString methods.
MySqlNativeTypeMetadata: Added isMariaDbJson method to check if the value of extendedTypeInfo equals 'json'.
MySqlType: In the MySqlType.of method, if type_id is ID_VARCHAR, ID_VAR_STRING or ID_STRING, returns VARBINARY if data is binary and if not returns JSON if isMariaDbJson() is true, otherwise returns VARCHAR.
DefinitionMetadataMessage: Added nullable extendedTypeInfo field. Added getter for this field and included it in hashcode, equals and toString methods. In decode41 method, if isMariaDb and isExtendedTypeInfo capabilities are enabled, added extendTypeInfo field that reads extended type information from buf.
MySqlRowDescriptorTest: Added null in MySqlColumnDescriptor argument to accommodate changes made to MySqlColumnDescriptor.
MySqlTypeMetadataTest: Added test method that checks if metadata isMariaDbJson is true/false depending on the value of extendedTypeInfo. Also checks if MYSQL_TYPE_STRING returns correct MySqlType (JSON/VARCHAR) depending on if extendedTypeInfo is 'json' or not.

Result:
If extended type information capability is enabled for MariaDb and a MariaDb JSON column is returned, the column should have a MySqlType of JSON rather than VARCHAR.
The drawbacks are that the extendedTypeInfo field has been added to several classes even though it's often null, which is always the case when using MySql and often the case when using MariaDb (only not null when extended type info capability is enabled and using protocol 4.1). Also even when it's not null, it only affects MySqlType when the value is 'json'.
Added a new test for MySqlTypeMetadata.


r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/message/server/DefinitionMetadataMessage.java
```diff
@@ -96,6 +100,10 @@ public int getDefinitions() {
     public short getDecimals() {
         return decimals;
     }
+
```
jchrys: `@Nullable`?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/message/server/DefinitionMetadataMessage.java
```diff
@@ -115,21 +123,22 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public String toString() {
         return "DefinitionMetadataMessage{database='" + database + "', table='" + table + "' (origin:'" +
-            originTable + "'), column='" + column + "' (origin:'" + originColumn + "'), collationId=" +
-            collationId + ", size=" + size + ", type=" + typeId + ", definitions=" + definitions +
-            ", decimals=" + decimals + '}';
+            originTable + "'), column='" + column + "' (origin:'" + originColumn + "'), extendedTypeInfo=" +
+        	extendedTypeInfo + ", collationId=" +collationId + ", size=" + size + ", type=" + typeId +
```
jchrys: Please wrap the string in single quotes `'`.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadata.java
```diff
@@ -117,20 +133,22 @@ public boolean equals(Object o) {
...
-        return typeId == that.typeId && definitions == that.definitions && collationId == that.collationId;
+        return typeId == that.typeId && definitions == that.definitions && collationId == that.collationId && 
+        		Objects.equals(extendedTypeInfo, that.extendedTypeInfo);
     }
 
     @Override
     public int hashCode() {
         int result = 31 * typeId + (int) definitions;
-        return 31 * result + collationId;
+        return 31 * result + collationId + extendedTypeInfo.hashCode();
```
jchrys: Looks like it's not null-safe. Could you make it null-safe and apply *31 as well?
ex)
```
        result =  31 * result + collationId;
        return 31 * result + (extendedTypeInfo == null ? 0 : extendedTypeInfo.hashcode)
```

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadata.java
```diff
@@ -117,20 +133,22 @@ public boolean equals(Object o) {
...
     }
 
     @Override
     public String toString() {
         return "MySqlTypeMetadata{typeId=" + typeId +
             ", definitions=0x" + Integer.toHexString(definitions) +
-            ", collationId=" + collationId +
+            ", collationId=" + collationId + 
+            ", extendedTypeInfo=" + extendedTypeInfo +
```
jchrys: single quote is missing (`'`)

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadata.java
```diff
@@ -105,6 +116,11 @@ public boolean isEnum() {
     public boolean isSet() {
         return (definitions & SET) != 0;
     }
+    
+    @Override
+    public boolean isMariaDbJson() {
+    	return extendedTypeInfo.equals("json");
```
jchrys: seems not null safe

r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadataTest.java
```diff
@@ -50,11 +52,24 @@ void noSet() {
...
+        MySqlTypeMetadata metadata = new MySqlTypeMetadata(0, -1, CharCollation.BINARY_ID, null);
 
         assertThat(metadata.isBinary()).isTrue();
 
-        metadata = new MySqlTypeMetadata(0, -1, 33);
+        metadata = new MySqlTypeMetadata(0, -1, 33, null);
         assertThat(metadata.isBinary()).isFalse();
     }
+    
+    @Test
```
jchrys, line 63: Could you add an integration test as well to make sure everything works?

jchrys (changes requested):
Thanks a lot! Just a small nit
(Also, could you check the result of the check-style plugin to ensure CI runs properly?)

svats0001:
Hi @jchrys ,

I appreciate the feedback and I'll incorporate those changes in a new commit as well as the checkstyle changes. I'm having some temporary issues running checkstyle in my Eclipse IDE so as soon as I figure that out I'll resubmit the changes in some time.

jchrys:
Hi, @svats0001.
Thanks for the update! No rush, just submit the changes when you're ready. :D

r2dbc-mysql/src/test/java/io/asyncer/r2dbc/mysql/MySqlTypeMetadataTest.java
svats0001: Hi @jchrys ,
I've added an integration test in MariaDbIntegrationTestSupport but I'm unsure of how to set/enable the extended type info capability for the test server. Do you know how to set this so that the test won't fail?

jchrys: Hello, @svats0001.
Extended type info was introduced in mariadb 10.5. I believe adding `@EnabledIf("envIsMariaDb10_5_1")` would accommodate this.

svats0001: Hmm, seems like the database isn't returning the extended type information @jchrys . Is there some flag that we need to set for the database to return this info or should it automatically return this after version 10.5 (this is what I assume to be true)?

svats0001:
Hi @jchrys ,
I think it's ready for review now, I was having some issues understanding how capabilities work but I think I understand now.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/Capability.java
```diff
@@ -169,7 +169,11 @@ public final class Capability {
...
 //    private static final long MARIADB_CLIENT_STMT_BULK_OPERATIONS = 1L << 34;
-//    private static final long MARIADB_CLIENT_EXTENDED_TYPE_INFO = 1L << 35;
+
+    /**
+     * Receive extended column type information from MariaDB to find out more specific details about column type.
+     */
+    private static final long MARIADB_CLIENT_EXTENDED_TYPE_INFO = 1L << 35;
 //    private static final long MARIADB_CLIENT_CACHE_METADATA = 1L << 36;
 
     private static final long ALL_SUPPORTED = CLIENT_MYSQL | FOUND_ROWS | LONG_FLAG | CONNECT_WITH_DB |
```
jchrys, line 179: When I add the `MARIADB_CLIENT_EXTENDED_TYPE_INFO` capability to `ALL_SUPPORTED`, I get an error. Could you check this out?

jchrys (commented):
@svats0001
Thanks for all your hard work! I was doing a final check, and I encountered an error when I enabled it. Could you take a look?

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/Capability.java
jchrys:
```java
private static final long ALL_SUPPORTED = CLIENT_MYSQL | FOUND_ROWS | LONG_FLAG | CONNECT_WITH_DB |
                                              NO_SCHEMA | COMPRESS | LOCAL_FILES | IGNORE_SPACE | PROTOCOL_41 |
                                              INTERACTIVE | SSL |
                                              TRANSACTIONS | SECURE_SALT | MULTI_STATEMENTS | MULTI_RESULTS |
                                              PS_MULTI_RESULTS |
                                              PLUGIN_AUTH | CONNECT_ATTRS | VAR_INT_SIZED_AUTH | SESSION_TRACK |
                                              DEPRECATE_EOF | ZSTD_COMPRESS | MARIADB_CLIENT_EXTENDED_TYPE_INFO
```

svats0001: When I was testing it the MariaDB versions >= 10.5 didn't seem to return the extended_type_info capability which is what confused me as I manually enabled these capabilities in the test and it didn't work because the packet didn't contain the extended information. In this case the capability is enabled by default but the packet doesn't contain the extended information so if anything you shouldn't include the extended_type_info capability in ALL_SUPPORTED and instead rely on the server returning the capability if enabled. But then the capability should have been turned off if not enabled by the server should it not..

jchrys: Oh, I understand. In my setup(and github), [the capability variable](https://github.com/asyncer-io/r2dbc-mysql/blob/acff42903b2a54a8d526f4ec2c7fc38c397fc567/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/message/server/HandshakeV10Request.java#L138-L138) represents the server-side capability, which is provided directly by the server. This capability includes extended_type_info, but the capability.extendMariaDb line [here](https://github.com/asyncer-io/r2dbc-mysql/blob/acff42903b2a54a8d526f4ec2c7fc38c397fc567/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/message/server/HandshakeV10Request.java#L145-L145) removes this flag due to the influence of [ALL_SUPPORTED](https://github.com/asyncer-io/r2dbc-mysql/blob/acff42903b2a54a8d526f4ec2c7fc38c397fc567/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/Capability.java#L380-L380).

jchrys: @svats0001 Could you try using MariaDB version 11.5.2? Alternatively, you could run the test with the VM options `-Dtest.db.type=mariadb -Dtest.db.version=11.5.2`. The official Docker image supports the flag by default.

svats0001: Oh I overlooked that fact about ALL_SUPPORTED, that makes sense.

I've tried with version 11.5.2 as well and it's not working either. The fact that all the tests run fine without trying to read extended_type_info suggests that the column definition packets don't contain this information. In the docs for extended_type_info, it does say 'while string has data' which may mean that it only returns the data for specific columns, so I'll play around with that some more by testing conditionally on the value of the first byte. Also, even versions < 10.5 return the extended_type_info flag because the tests fail for these versions too which is weird.

svats0001: It looks like I'm not able to progress this issue any further, I'll close this pull request if that's okay @jchrys

jchrys: @svats0001 

Would it be possible to keep this PR open so that I can continue working on it when I have some availability? I believe this work is valuable and deserves to be delivered.

svats0001: Yep no problem, I'll leave it open.

svats0001:
@jchrys Closing this and making new one.

Issue #285, opened 2024-07-29
Implement Relaxed Conversion Logic for Codecs


jchrys:
Existing codec implementation exhibits strict type conversions. For instance, boolean codec converts to boolean only when data type is binary or TINYINT(1) BIT(1). However, In many real-world applications, data stored in MySQL databases can be represented in multiple formats. Specifically, Boolean values might be stored in columns with different data types such as VARCHAR or CHAR, where values like 'true', 'false', '1', '0' are used to represent Boolean states.  The current implementation does not support converting these string representations to Boolean values, causing inconvenience and additional data transformation efforts.

jchrys:
ref: type conversions
https://dev.mysql.com/doc/connector-j/en/connector-j-reference-type-conversions.html

PR #290, opened 2024-10-21
Implemented relaxed conversion logic for BooleanCodec

svats0001:
@jchrys #285 

Motivation:
To allow boolean values stored as strings in MySQL database such as "true", "false", "1" and "0" to be converted to their corresponding boolean values.

Modification:
BooleanCodec: Changed decode method to check if VARCHAR value is a boolean value. Changed doCanDecode to add VARCHAR.
BooleanCodecTest: Added decodeString test to ensure boolean values stored as strings are converted into the correct corresponding boolean values.

Result:
Boolean values stored as strings can now be converted to their corresponding boolean values. Drawbacks are that there could be a string column containing numeric data with values other than 0 or 1 and the column isn't used for storing boolean values at the same time the codec interprets the 0's and 1's as boolean. Only boolean values "true", "false", "1" and "0" are decoded, other possible types of boolean value strings haven't been included.  Also, doCanDecode states that the VARCHAR data type can be decoded but only a small subset of this data type can be decoded and it's not possible to highlight the conditions in the doCanDecode method.

### DIFF
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/codec/BooleanCodec.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/codec/BooleanCodec.java
@@ -24,6 +24,7 @@ import io.asyncer.r2dbc.mysql.api.MySqlReadableMetadata;
 import io.asyncer.r2dbc.mysql.constant.MySqlType;
 import io.netty.buffer.ByteBuf;
 import io.netty.buffer.ByteBufAllocator;
+import io.r2dbc.spi.R2dbcNonTransientResourceException;
 import reactor.core.publisher.Mono;
 
 /**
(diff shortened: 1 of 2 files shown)


jchrys:
Thanks! To closely align with MySQL Connector/J's behavior, please refer to the [BooleanValueFactory logic here](https://github.com/mysql/mysql-connector-j/blob/7d6b0800528b6b25c68b52dc10d6c1c8429c100c/src/main/core-impl/java/com/mysql/cj/result/BooleanValueFactory.java#L92-L116). Additionally, avoiding fallback on other types' decoding logic can help minimize complexity
If numeric handling is challenging, starting with non-numeric cases could be a good first step.

svats0001:
Hi @jchrys ,
I've altered the method so that it conforms with MySQL Connector/J's behaviour. I'm not sure exactly which type of exception needs to be thrown though (if a new type of exception needs to be created), or if an exception needs to be thrown at all.

jchrys:
@svats0001 
It looks great. let me go through it :D

jchrys:
@svats0001
Could you include test cases using scientific notation, such as 1e4 and -1.34e10, in the tests? Also, please include -0. Everything else looks good to me.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/codec/BooleanCodec.java
```diff
@@ -38,7 +40,33 @@ private BooleanCodec() {
...
+                return createFromLong(0);
+            } else if (s.matches("-?\\d*\\.\\d*") || s.matches("-?\\d*\\.\\d+[eE]-?\\d+")) {
+                return createFromDouble(Double.parseDouble(s));
+            } else if (s.matches("-?\\d+")) {
+                if (!CodecUtils.isGreaterThanLongMax(s)) {
+                    return createFromLong(CodecUtils.parseLong(value));
+                }
+                return createFromBigInteger(new BigInteger(s));
+            }
+            throw new IllegalArgumentException("Unable to interpret string: " + s);
```
jchrys: For this type of error, `R2dbcNonTransientResourceException` is appropriate. Could you also include the type in the message? like "The value '%s' of type '%s' cannot be decoded into a Boolean." with sqlState: '22018'

svats0001:
@jchrys , I've added those tests and changed the exception now.

jchrys (approved):
lgtm
@svats0001 Thanks a lot

Issue #335, opened 2026-03-22
[bug] SSL VERIFY_CA fails with Hostname or IP address is undefined on Spring Boot 4


underbell:
**Describe the bug**
When using `sslMode=VERIFY_CA` with `r2dbc-mysql` on Spring Boot 4, the connection fails with:
`java.security.cert.CertificateException: Hostname or IP address is undefined`.

The same connection works correctly when using `sslMode=REQUIRED`.

This suggests that hostname verification is being triggered unexpectedly in `VERIFY_CA` mode.

**To Reproduce**
1. Use the following environment:
   - `io.asyncer:r2dbc-mysql:1.4.1`
   - Spring Boot 4.x
2. Configure connection with:
   - `r2dbc:mysql://host:3306/db?sslMode=VERIFY_CA&sslCa=/path/to/ca-cert.pem` 
3. Attempt to connect
4. Observe the following error:
```
Caused by: java.security.cert.CertificateException: Hostname or IP address is undefined.
at sun.security.util.HostnameChecker.match(...)
at sun.security.ssl.X509TrustManagerImpl.checkIdentity(...)
```
5. Change configuration to: sslMode=REQUIRED
→ Connection succeeds

**Expected behavior**
Connection succeeds

jchrys:
@underbell  
Thanks for the report
Fixed in #340 and released in 1.4.2. Please give it a try.

underbell:
Thanks. I was glad that the temporary code I wrote was the same as the code that resolved the issue. I was really worried about whether there was a better way. I'll try to contribute next time.

jchrys:
Great. Your workaround was spot on. :D  Contributions are always welcome.

PR #337, opened 2026-04-20
Fix caching_sha2_password full auth over Unix socket

lorenz1980:
## Motivation

`caching_sha2_password` full authentication currently fails over a Unix domain socket connection because `HandshakeExchangeable` unconditionally requires SSL whenever the auth provider's `isSslNecessary()` returns `true`. Over Unix sockets this rejects the client before any request is sent.

MySQL's own documentation explicitly treats Unix sockets as a secure transport for `caching_sha2_password`:

> If the connection is secure, an RSA key pair is unnecessary and is not used. This applies to TCP connections encrypted using TLS, as well as Unix socket-file and shared-memory connections.
> — [MySQL 8.4 Reference — Caching SHA-2 Pluggable Authentication](https://dev.mysql.com/doc/refman/8.4/en/caching-sha2-pluggable-authentication.html)

Other drivers already handle this correctly:
- MySQL Connector/J
- `libmysqlclient` (the reference C client)
- `go-sql-driver/mysql`
- Similar issue tracked in `aiomysql`: [#673](https://github.com/aio-libs/aiomysql/issues/673)

This becomes blocking on MySQL 9.0, which removes `mysql_native_password` entirely. Any Unix-socket deployment must use `caching_sha2_password` — and today that combination does not work with `r2dbc-mysql`.

## Modification

Conceptually: the existing `sslCompleted` check already models "the transport is already secure, so plaintext auth is safe." Unix sockets are another form of secure transport. We generalise the existing check rather than introducing new concepts.

- **`Client` interface** — add `default boolean isLocalTransport()` returning `false`. Default-method, non-breaking for any third-party `Client` implementations.
- **`ReactorNettyClient`** — override; query the live Netty channel's `remoteAddress()` for `io.netty.channel.unix.DomainSocketAddress`. No new state stored; the address is already encoded in the channel.
- **`HandshakeExchangeable`** — extend the SSL-requirement checks at the three sites in `createAuthResponse` / `createHandshakeResponse` so that a local transport short-circuits the SSL requirement in the same way a completed TLS handshake already does.
- **No changes to `MySqlAuthProvider` or any auth provider implementations.** The auth method's wire format is unchanged; only the decision of whether this channel is secure enough for that wire format moves.

No public API break. No behaviour change for TCP connections (with or without TLS).

## Result

- `caching_sha2_password` full authentication succeeds over Unix domain sockets without TLS.
- TCP without TLS still rejects (existing behaviour preserved).
- TLS path unchanged.
- RSA-encryption path (added in `a14da54`) unchanged for TCP; skipped for Unix socket since plaintext is already safe.

### Tests

New unit test `InitFlowUnixSocketTest` (matches existing unit-test style — JUnit 5 + Mockito + AssertJ, same package, reflection to reach `private createAuthResponse`):

1. Full auth over Unix socket, SSL disabled → succeeds (the reported bug, fixed).
2. Full auth over TCP without TLS → throws `R2dbcPermissionDeniedException` (existing behaviour preserved).
3. Full auth with TLS completed → succeeds (existing TLS path preserved).

All 437 existing unit tests continue to pass (no regressions).

### Not included (intentionally, to keep PR scope tight)

- Integration test via `TestContainerExtension`. The default testcontainer (`mysql:5.7.44`) doesn't use `caching_sha2_password` by default, and bind-mounting the container's Unix socket to the host on CI adds platform-specific complexity (macOS native transport, ARM/x86). The fix has been verified end-to-end locally with a MySQL 8.4 container + Unix socket + `caching_sha2_password` + cold cache — happy to add a testcontainer-based integration test in a follow-up if maintainers request it.
- De-duplicating the two `instanceof` branches (the existing one in `Client.connect()` for TCP options, and the new one in `ReactorNettyClient.isLocalTransport()`). Both read the live address; neither stores state, so they cannot drift. Happy to extract a shared helper in a follow-up.

### DIFF
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConnectionContext.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/ConnectionContext.java
@@ -55,6 +55,8 @@ public final class ConnectionContext implements CodecContext {
 
     private final boolean preserveInstants;
 
+    private final boolean unixSocket;
+
     private int connectionId = -1;
 
     private ServerVersion serverVersion = NONE_VERSION;
@@ -111,7 +113,8 @@ public final class ConnectionContext implements CodecContext {
         int localInfileBufferSize,
         boolean tinyInt1isBit,
         boolean preserveInstants,
-        @Nullable ZoneId timeZone
+        @Nullable ZoneId timeZone,
+        boolean unixSocket
     ) {
         this.zeroDateOption = requireNonNull(zeroDateOption, "zeroDateOption must not be null");
         this.localInfilePath = localInfilePath;
@@ -119,6 +122,18 @@ public final class ConnectionContext implements CodecContext {
         this.tinyInt1isBit = tinyInt1isBit;
         this.preserveInstants = preserveInstants;
         this.timeZone = timeZone;
+        this.unixSocket = unixSocket;
+    }
+
+    /**
+     * Checks whether the underlying transport is a Unix domain socket, which MySQL treats as a secure channel
+     * and therefore permits {@code caching_sha2_password} full authentication to send the password in
+     * plaintext without TLS, matching the behaviour of {@code libmysqlclient} and other drivers.
+     *
+     * @return {@code true} if connected over a Unix domain socket; {@code false} otherwise.
+     */
+    boolean isUnixSocket() {
+        return unixSocket;
     }
 
     /**
--- a/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/InitFlow.java
+++ b/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/InitFlow.java
@@ -611,7 +611,7 @@ final class HandshakeExchangeable extends FluxExchangeable<Void> {
     private AuthResponse createAuthResponse(String phase) {
         MySqlAuthProvider authProvider = getAndNextProvider();
 
-        if (authProvider.isSslNecessary() && !sslCompleted) {
+        if (authProvider.isSslNecessary() && !sslCompleted && !client.getContext().isUnixSocket()) {
             if (serverRSAPublicKeyFile != null && sslMode.equals(SslMode.DISABLED)) {
                 return new AuthResponse(MySqlAuthProvider.rsaEncryption(authProvider.authentication(
                     password, salt, client.getContext().getClientCollation()), serverRSAPublicKeyFile,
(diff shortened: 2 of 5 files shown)


lorenz1980:
Hi @jchrys 👋 — gentle nudge on this one, since it's assigned to you.

The PR is rebased onto current `trunk` and the full CI matrix is green: unit tests on Java 8/11/17/21, and integration tests across MySQL 5.5–9.2 and MariaDB 10.0–10.11, plus CodeQL, GraalVM, and DCO.

Of particular note, **MySQL 9.0 passes** — which is the case this fix specifically addresses: with `mysql_native_password` removed in 9.0, `caching_sha2_password` over a Unix domain socket previously couldn't complete auth at all. The change treats a local (Unix socket) transport as secure for the SSL-requirement check, matching how Connector/J, libmysqlclient, and go-sql-driver already behave. It's covered by `InitFlowUnixSocketTest`.

It's orthogonal to your #335 endpoint-identification work (different code path — no TLS involved), so they should sit together cleanly.

Would you have a chance to take a look when time permits? Happy to adjust anything. Thanks!

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/client/Client.java
```diff
@@ -107,6 +107,19 @@ public interface Client {
...
     boolean isConnected();
 
+    /**
+     * Checks whether the underlying transport is a local transport (e.g. Unix domain socket) that MySQL treats
+     * as a secure channel.  When this is {@code true}, {@code caching_sha2_password} full authentication is
+     * allowed to send the password in plaintext without TLS, matching the behaviour of
+     * {@code libmysqlclient} and other drivers.
+     *
+     * @return {@code true} if the transport is local (e.g. Unix domain socket); {@code false} otherwise
+     * @since 1.4.2
```
jchrys:
```suggestion
     * @since 1.4.3
```

jchrys (commented):
LGTM, just a nit

lorenz1980:
Thanks @jchrys! Applied — bumped to `@since 1.4.3` in `96ba338`. Let me know if there's anything else.

r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/client/Client.java
```diff
@@ -107,6 +107,19 @@ public interface Client {
...
+     */
+    default boolean isLocalTransport() {
```
jchrys: Can we keep this out of the public API? I think we could add [`isUnixSocket()` ](https://github.com/asyncer-io/r2dbc-mysql/blob/adfb4282923f6b221bf74c98561bb7c29a0e506e/r2dbc-mysql/src/main/java/io/asyncer/r2dbc/mysql/MySqlConnectionFactoryProvider.java#L68)as a flag on `ConnectionContext` instead. 

for example: 
```java
// ConnectionContext.java
private final boolean unixSocket;
boolean isUnixSocket();
new ConnectionContext(..., !configuration.isHost()); // WDYT?
```

lorenz1980:
Done in `3150fdb` — moved it off the public `Client` API onto `ConnectionContext`:

- Added a package-private `ConnectionContext#isUnixSocket()`, set at construction via `new ConnectionContext(..., !configuration.isHost())`, exactly as you suggested.
- Removed `Client#isLocalTransport()` and its `ReactorNettyClient` override.
- The three SSL-requirement gates in `InitFlow` now read `!client.getContext().isUnixSocket()`.
- Updated `InitFlowUnixSocketTest` / `ConnectionContextTest` accordingly (both green locally).

For the record on why it was originally on `Client`: the gate is `!sslCompleted && !isLocalTransport()`, and I'd modeled both as live-channel "is the wire already secure?" facts for symmetry. But since the `DomainSocketAddress` is derived 1:1 from `configuration.isHost()` in `MySqlConnectionFactory`, the config-level flag carries identical information without widening the public API — so your version is cleaner. Thanks for the review!

jchrys (approved):

jchrys:
@lorenz1980  thanks a lot! 
One last thing before merge, please update the commit message. It still describes the old approach.

lorenz1980:
Updated the commit message in `bb0ffbb` to describe the `ConnectionContext#isUnixSocket()` approach (code unchanged — message-only amend). Thanks again for the review!

jchrys:
@lorenz1980 Thanks a lot :+1:

